PR #2, opened 2017-08-07
Fmt

squat:
Add a basic Makefile for checking and formatting TF files.


dghubble:
Ah, the formatting was addressed already, but thanks

Issue #13, opened 2017-09-09
Deprecate control plane self-hosted etcd


ericchiang:
### Problem

Self-hosted etcd is still marked as experimental by bootkube, and the actual pivot between the on host etcd and self-hosted etcd a significant chunk of the startup time for bootkube clusters (~20%). Anecdotally, I've also seen self-hosted clusters fail, which was part of the inspiration for the `bootkube recover` sub-command.

### Desired Behavior

Support on-host etcd clusters. Possibly either by provisioning separate machines, or having the etcd instances living on the masters. Etcd instances on the master nodes has an attractive aspect of reduced overhead, but might not be optimal due to masters living in an auto-scaling group.

### Tradeoffs

Etcd doesn't behave well in an auto-scaling group.[1] This might be alleviated by talking to the cloud provider or not supporting expanding the number of masters, but is worth investigation on its own. Bootstrapping needs a concrete, cross cloud solution, and re-balancing must survive reboots and adding new members after the initial cluster is formed.

A large chunk of this feature is determining how best to bootstrap and manage the etcd cluster.

[1] https://crewjam.com/etcd-aws/

ericchiang:
I'd like to do some initial exploration, so please feel free to assign me for now.

dghubble:
The self-hosted etcd problems are outlined in FAQ https://typhoon.psdn.io/faq/#self-hosted-etcd The plan is to revert back to etcd on controllers (not an instance group), which is how we were doing it before self-hosted etcd came along.

dghubble:
I've dropped self-hosted etcd as an option on bare-metal and will begin removing it from cloud clusters. In the mean time, I've had one AWS (alpha) cluster crash related to self-hosted etcd.

https://github.com/coreos/etcd-operator/issues/731
https://github.com/kubernetes/kubernetes/issues/40973

dghubble:
Digital Ocean now uses on-host (i.e. on-controller) etcd. #32

dghubble:
AWS now uses on-host etcd. #53 

All platforms now use on-host etcd across controllers. Self-hosted etcd is no longer used. Clusters should be more stable, reliable, and bootstrap times are much nicer (< 6min).

PR #32, opened 2017-10-04
digital-ocean: Run etcd cluster on-host, across controllers

dghubble:
* Run etcd peers with TLS across controller nodes
* Deprecate self-hosted-etcd on the Digital Ocean platform
* Distribute etcd TLS certificates as part of initial provisioning
* Check the status of etcd by running `systemctl status etcd-member`
* Side effect: Cuts down provisioning time from 5:05 to 3:30.

rel: #13

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,14 @@
 
 Notable changes between versions.
 
+## Latest
+
+#### Digital Ocean
+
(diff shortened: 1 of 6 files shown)


dghubble:
This change needs to simmer for a bit.

PR #47, opened 2017-10-26
Update from docker 1.12 to docker 17.09

dghubble:
Kubernetes 1.8 recommends docker 17.03. Container Linux stable ships with 1.12 which is still fine for Kubernetes 1.8. An upcoming stable release will switch to 17.09 though. This PR shares some docker 17.09 testing work I've been doing to help identify any issues.

**UPDATE: Typhoon will move directly to Docker 17.09.**

Docker's version mess:

* 1.12 - technically (near)deprecated by Docker. Most well used since most Container Linux Kubernetes clusters use this version. 
* 17.03 - Kubernetes 1.8 recommendation
* 17.09 - Upcoming Docker version on Container Linux stable

So far, clusters run ok with Docker 17.09.
```
$ docker version
Client:
 Version:      17.09.0-ce
 API version:  1.32
 Go version:   go1.8.2
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:24:58 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.09.0-ce
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.2
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:24:58 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

```
$ kubectl describe nodes | grep docker
Container Runtime Version:     docker://Unknown
...
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,10 +5,9 @@ Notable changes between versions.
 ## Latest
 
 * Kubernetes v1.8.4
-* Calico related bug fixes
+* Enable Docker 17.09 on Container Linux (replaces Docker 1.12)
 * Update Calico from v2.6.1 to v2.6.3
 * Update flannel from v0.9.0 to v0.9.1
-* Use service accounts for kube-proxy and pod-checkpointer
 
 ## v1.8.3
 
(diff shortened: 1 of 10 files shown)


dghubble:
CoreOS beta and alpha already default to docker 17.09, while CoreOS stable doesn't yet.

This change opts-in CoreOS stable to install 17.09. However, the way that is done, the stable image still comes with docker 1.12, but 17.09 is installed, which adds noticeably to the cluster creation time. We may be better off just doing nothing and waiting for the upcoming stable that switches to 17.09 as part of an auto-update.

dghubble:
The latest Container Linux images (stable, beta, and alpha) all default to docker 17.09, which works well with Typhoon.

Issue #57, opened 2017-11-08
Flexvolume volume-plugin-dir is read-only on coreos


bcg62:
https://github.com/coreos/docs/issues/1172

This affects things like rook
https://rook.github.io/docs/rook/master/k8s-pre-reqs.html

dghubble:
Agreed, the kubelet flag and mount could be added across platforms. I've had to set it manually to evaluate rook so it'd be nice to standardize it and shouldn't cause any downsides for those not using volume plugins.

Calico and flannel have established a pattern where they use a sidecar pod to install needed binaries on hosts `/opt/cni/bin` (which happens to be the default ). Flex volume plugins are similar and I think special enough to merit the same treatment - sidecar installs to known location, kubelet mounts.

I'm a bit partial to continuing to use `/opt/name` for these sorts of cases, but `/var/lib/kubelet/volumeplugins` (from rook docs) isn't bad either.

It might be a nice exercise to ensure it works for rook and some other popular flex volume plugin too, though I've only used rook myself.

cc @barakmich

dghubble:
I'd like to include #61 on bare-metal or maybe all platforms in the next release. I'm also not a heavy rook user at the moment, any chance you've looked at the change or have opinions? Which platform(s) are you seeking?

bcg62:
lgtm, my use case at the moment is bare metal

dghubble:
Added on bare-metal. I'll verify with rook soon, but these flags are working and sensible to always have enabled.

ivan4th:
This problem is not limited to bare-metal. For example, I want to use [DigitalOcean block storage](https://github.com/kubernetes-incubator/external-storage/tree/master/digitalocean) via `external-storage`, and to do so I also need to be able to install a flexvolume plugin.

ivan4th:
In addition to that, DigitalOcean block storage plugin executable needs to be accessible by kube-controller-manager, it's possible to achieve that by for example mounting `/var/lib/kubelet/volumeplugins` from the host under `/usr/libexec/kubernetes/kubelet-plugins/volume/exec/` in the kube-controller-manager pod. I can try and make a PR to support it but I'm not sure what's the best way to add mounts to the controller manager pod.

dghubble:
I have no objection to providing the `--volume-plugin-dir=/var/lib/kubelet/volumeplugins` flag to the kubelet on cloud platforms, it corrects the search path for Container Linux. It was simply added to bare-metal first.

[...]

PR #61, opened 2017-11-11
Add kubelet --volume-plugin-dir flag on bare-metal

dghubble:
* Kubelet will search path for flexvolume plugins

rel: #57


barakmich:
LGTM; I've written this one a few times. It may be useful on the other platforms, if only for consistency?

PR #68, opened 2017-12-02
Remove version constraints on provider plugins

dghubble:
* Terraform v0.11.x strongly discourages modules which have provider blocks and notably doesn't delete modules correctly if they are present https://github.com/hashicorp/terraform/issues/16824. Not being able to delete clusters is a show-stopper.
* Unfortunately, this means there are no safeguards against users continuing to use old or incompatible provider plugins
* https://github.com/hashicorp/terraform/issues/16777
* https://github.com/hashicorp/terraform/issues/16824

### DIFF
--- a/aws/container-linux/kubernetes/require.tf
+++ b/aws/container-linux/kubernetes/require.tf
@@ -1,25 +1,5 @@
-# Terraform version and plugin versions
+# Terraform version
 
 terraform {
   required_version = ">= 0.10.4"
-}
-
-provider "aws" {
-  version = "~> 1.0"
-}
-
-provider "local" {
-  version = "~> 1.0"
-}
-
-provider "null" {
-  version = "~> 1.0"
-}
-
-provider "template" {
-  version = "~> 1.0"
-}
-
-provider "tls" {
-  version = "~> 1.0"
 }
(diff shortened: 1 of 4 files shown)


dghubble:
Based on conversation in https://github.com/hashicorp/terraform/issues/16824, to allow users to use Terraform v0.11.x, we don't have to remove these version constraints. But we do need to add a changelog/break/documentation for the new format:

Copy providers into a top level `providers.tf`:

```
provider "local" {
  version = "~> 1.0"
  alias = "default"
}

provider "null" {
  version = "~> 1.0"
  alias = "default"
}

provider "template" {
  version = "~> 1.0"
  alias = "default"
}

provider "tls" {
  version = "~> 1.0"
  alias = "default"
}
```

And also edit every module usage to have a `providers` block:

```
module "aws-cluster" {
  source = "git::https://github.com/poseidon/typhoon//aws/container-linux/kubernetes?ref=5ea7ce0af559857591f20fe19b03aab177fd7032"
  
  providers = {
    aws = "aws.default"
    local = "local.default"
    null = "null.default"
    template = "template.default"
    tls = "tls.default"
  }

  cluster_name = "blah"
   ...
```

CamelCaseNotation:
How exactly are you using the `providers.tf` file?

PR #69, opened 2017-12-04
Fixes issue with etcd DNS resolution via locally provided nameserver.

shift:
## Issue

When restarting masters, `etcd-member.service` fails to be able to reverse lookup the names of the TLS nodes as DNS hasn't been defined yet. As a side effect causes some issues with CLUO deployed ;)

```
Dec 04 07:27:16 node0.cluster.com etcd-wrapper[875]: 2017-12-04 07:27:16.092239 I | etcdmain: rejected connection from "192.168.15.13:44308" (tls: "192.168.15.13" does not match any of DNSNames ["node0.int.cluster.com" "node3.int.cluster.com" "*.kube-etcd.kube-system.svc.cluster.local" "kube-etcd-client.kube-system.svc.cluster.local"])
```

I've only used this on bare-metal so not sure if it effects other module types.

## Change Details

This change makes it so that the `wait-for-dns.service` uses the `RequiredBy` directive for `etcd-member.service`.

## Testing

I've used this fix previously from prior to Typhoon and has been running successfully in more than one cluster.

### DIFF
--- a/aws/container-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/aws/container-linux/kubernetes/cl/controller.yaml.tmpl
@@ -41,6 +41,7 @@ systemd:
         ExecStart=/bin/sh -c 'while ! /usr/bin/grep '^[^#[:space:]]' /etc/resolv.conf > /dev/null; do sleep 1; done'
         [Install]
         RequiredBy=kubelet.service
+        RequiredBy=etcd-member.service
     - name: kubelet.service
       enable: true
       contents: |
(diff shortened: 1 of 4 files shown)


dghubble:
Thanks! I've seen this race occur on rare occasions, but didn't get to changing it. If etcd-member.service starts before `/etc/resolv.conf` is populated the container is left with an empty `/etc/resolv.conf` and cannot resolve its peers. The quick fix is to `systemctl restart etcd-member.servce`.

Its unfortunate, but I think having both kubelet and now etcd require the little wait-for-dns shim unit is a suitable fix. It can occur on any platform so the fix should be applied in all controller container linux configs. Want to tweak the others?

shift:
Updated PR to include all 4 profiles :)

dghubble:
Thanks!

dghubble:
Merged as ce49a93d5d82b4b59863ee1ce05698c14fd0efe1 to normalize commit message

PR #70, opened 2017-12-07
Fix moving manifests in bootkube-start to work with multiple directories

elemental-lf:
## Issue

In /opt/bootkube/bootkube-start when there is more than one directory matching /opt/bootkube/assets/manifests-* the manifests won't be moved to /opt/bootkube/assets/manifests at all because -d fails with an error. The -d test only allows one argument.

Currently this works because there's only one directory (manifests-networking). But it seems that the original intention was to move the files in all matching directories. I'm also in need of this change because I'm working on some patches to add canal networking to terraform-bootkube. These changes split up the manifests-networking directory into multiple ones (because of the way the Terraform template_dir resource works) to share as much as possible between the Canal and Calico/flannel networking modes.

## Change details

This change replace the -d test with a call to ls to check for the existence of zero or more files matching the /opt/bootkube/assets/manifests-*/* pattern.

## Testing

This change has only been tested on bare-metal and works as expected.

### DIFF
--- a/aws/container-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/aws/container-linux/kubernetes/cl/controller.yaml.tmpl
@@ -147,7 +147,7 @@ storage:
           # Wrapper for bootkube start
           set -e
           # Move experimental manifests
-          [ -d /opt/bootkube/assets/manifests-* ] && mv /opt/bootkube/assets/manifests-*/* /opt/bootkube/assets/manifests && rm -rf /opt/bootkube/assets/manifests-*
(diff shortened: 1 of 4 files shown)


dghubble:
> it seems that the original intention was to move the files in all matching directories

Yeah, that was the intention if additional categories of optional manifests were ever needed. No desire to build out flannel (with canal), but this change is generally useful.

The subshell isn't the nicest, but the alternatives look worse and this works alright. Checked with missing match as well to ensure it correctly doesn't attempt to mv and rm. Going to do a little more checking and merge. Can you please normalize the commit message if you get a chance?

elemental-lf:
Thank you for looking into this. I've changed the commit message. Didn't know if I should include a subsystem (or "*:") in the summary or not as this wasn't handled consistently in the past. As this affects all platforms I went without one.

dghubble:
I rebased one more time just so the commit is nicely at the top eea79e895dac11092ea3fbecfbc13ea12fa32822. Thanks Lars.

PR #74, opened 2017-12-09
Support Bare Metal with custom cluster domains

barakmich:
Removes the restriction that all clusters must be in the FQDN `cluster.local` and adds an optional variable for a custom domain. `cluster.local` is still the default, so the net change should be fully backwards compatible 

* Adds variable for regular bare-metal and pxe clusters, passes it to everywhere it's required

## Testing

`terraform plan` produces reasonable configs

Depends on poseidon/terraform-render-bootkube#15


dghubble:
It seemed reasonable to use `cluster_domain_suffix` as mentioned in https://github.com/poseidon/terraform-render-bootkube/pull/15#issuecomment-351312387. So probably some sed replaces are in order, sorry. Rebase and sqash as well? Thanks for getting the docs too.

barakmich:
Resolved. If you want the squash, Github ought to give you the squash :)

PR #76, opened 2017-12-12
AWS Cloud Provider

valibud:
I wanted to be able to dynamically provision ebs volumes and create elb v1 from kubernetes service with type=LoadBalancer.

* Tag resources with cluster name,
* Ability to select CoreOS version,
* Set FQDN of CoreOS instance,
* Created IAM role and policy required for controller nodes.

## Testing

* I have deployed a cluster in AWS. Created a storageClass. Created statefulset with persistentVolumeClaims. EBS volumes created with success in AWS.

* I have created the following service with `kubectl apply`

```
kind: Service
apiVersion: v1
metadata:
  name: nginx-ingress-controller
  namespace: ingress-nginx
  labels:
    app: ingress-nginx
  annotations:
    # Enable PROXY protocol
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
    # Increase the ELB idle timeout to avoid issues with WebSockets or Server-Sent Events.
    service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '3600'
    # enable cross-zone load balancing
    service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
spec:
  type: LoadBalancer
  externalTrafficPolicy: Local
  healthCheckNodePort: 30000
  selector:
    app: ingress-nginx
  ports:
  - name: http
    # protocol: TCP
    port: 80
    targetPort: http
  - name: https
    protocol: TCP
    port: 443
    targetPort: https
  - name: healthz
    protocol: TCP
    port: 30000
    targetPort: 10254
```

TCP ELB created with success in AWS. Deploying nginx-ingress-controller and creating an ingress works as well.

### DIFF
--- a/aws/container-linux/kubernetes/ami.tf
+++ b/aws/container-linux/kubernetes/ami.tf
@@ -14,6 +14,6 @@ data "aws_ami" "coreos" {
 
   filter {
     name   = "name"
-    values = ["CoreOS-${var.os_channel}-*"]
+    values = ["CoreOS-${var.os_channel}-${var.os_version}*"]
   }
 }
--- a/aws/container-linux/kubernetes/bootkube.tf
+++ b/aws/container-linux/kubernetes/bootkube.tf
@@ -2,6 +2,7 @@
 module "bootkube" {
   source = "git::https://github.com/poseidon/terraform-render-bootkube.git?ref=533e82f833c166297abd249ac3d4853d6ebed364"
 
+  cloud_provider = "${var.cloud_provider}"
   cluster_name = "${var.cluster_name}"
   api_servers  = ["${format("%s.%s", var.cluster_name, var.dns_zone)}"]
   etcd_servers = ["${aws_route53_record.etcds.*.fqdn}"]
--- a/aws/container-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/aws/container-linux/kubernetes/cl/controller.yaml.tmpl
@@ -28,6 +28,21 @@ systemd:
       enable: true
     - name: locksmithd.service
       mask: true
+    - name: set-aws-fqdn.service
+      enable: true
+      contents: |
+        [Unit]
+        Wants=network-online.target
+        After=network-online.target
+        Description=Set Hostname Workaround
+        Before=kubelet.service
+
+        [Service]
+        Type=oneshot
+        ExecStart=/bin/sh -c "/usr/bin/hostnamectl set-hostname $(curl -s http://169.254.169.254/latest/meta-data/hostname)"
+
+        [Install]
+        RequiredBy=kubelet.service
     - name: wait-for-dns.service
       enable: true
       contents: |
@@ -74,6 +89,7 @@ systemd:
           --client-ca-file=/etc/kubernetes/ca.crt \
           --cluster_dns=${k8s_dns_service_ip} \
           --cluster_domain=cluster.local \
+          --cloud-provider=${cloud_provider} \
           --cni-conf-dir=/etc/kubernetes/cni/net.d \
           --exit-on-lock-contention \
           --kubeconfig=/etc/kubernetes/kubeconfig \
--- a/aws/container-linux/kubernetes/cl/worker.yaml.tmpl
+++ b/aws/container-linux/kubernetes/cl/worker.yaml.tmpl
@@ -5,6 +5,21 @@ systemd:
       enable: true
     - name: locksmithd.service
       mask: true
+    - name: set-aws-fqdn.service
+      enable: true
+      contents: |
+        [Unit]
+        Wants=network-online.target
+        After=network-online.target
+        Description=Set Hostname Workaround
+        Before=kubelet.service
+
+        [Service]
+        Type=oneshot
+        ExecStart=/bin/sh -c "/usr/bin/hostnamectl set-hostname $(curl -s http://169.254.169.254/latest/meta-data/hostname)"
+
+        [Install]
+        RequiredBy=kubelet.service
     - name: wait-for-dns.service
       enable: true
       contents: |
@@ -50,6 +65,7 @@ systemd:
           --client-ca-file=/etc/kubernetes/ca.crt \
           --cluster_dns=${k8s_dns_service_ip} \
           --cluster_domain=cluster.local \
+          --cloud-provider=${cloud_provider} \
           --cni-conf-dir=/etc/kubernetes/cni/net.d \
           --exit-on-lock-contention \
           --kubeconfig=/etc/kubernetes/kubeconfig \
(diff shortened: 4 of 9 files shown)


dghubble:
I left cloud-provider features out mainly due to the security ramifications. Yes, I'm aware commercial distros just ship it anyway, but its a rather unhappy tradeoff. I also wish the features weren't so all-or-nothing and could be isolated. EBS volume mounting is compelling, service type=LoadBalancer is a bit of an anti-pattern in my mind, Typhoon comes with appropriate Ingress infrastructure in place (and you don't pay your cloud provider $$$ for every app). Anyway, that's my ramble.

In practical terms, I'd like to get NetworkPolicy mature and enabled by default before this is considered.

Also:

- Can you explain the tags? I'm aware they're a "requirement" for `--cloud-provider=aws`, but I've yet to see a justification.
- Similar, for hostname setting requirement
- Please don't add a new variable for setting the Container Linux image version. Picking the stable, beta, or alpha channel should be sufficient and keeps things simple (also prevents self-foot injuries).

tsmith-pimco:
@dghubble without the `--cloud-provider=aws`, you cannot dynamically provision volumes which is a HUGE problem for the most of us. bootkube relies on the "KubernetesCluster" tag to find resources in aws, especially nodes in autoscaling groups.

taylorsmithgg:
@dghubble without the `--cloud-provider=aws`, you cannot dynamically provision volumes which is a HUGE problem for the most of us. bootkube relies on the "KubernetesCluster" tag to find resources in aws, especially nodes in autoscaling groups.

valentinbud:
Thanks for taking your time and look over my PR.

>  I also wish the features weren't so all-or-nothing and could be isolated.

I would like that too. As I see it now one way to achieve it would be with multiple `aws_iam_role_policy` and a flag to activate for example ebs volumes, elb, etc.

 >  EBS volume mounting is compelling, service type=LoadBalancer is a bit of an anti-pattern in my mind, Typhoon comes with appropriate Ingress infrastructure in place (and you don't pay your cloud provider $$$ for every app). Anyway, that's my ramble.

I think it's great Typhoon comes with appropriate Ingress infrastructure in place. It helped me understand way faster how thing work. And the fact you have it up and running in no time it's amazing. 

I also agree about service type=LoadBalancer being an anti-pattern. Having aws cloud provider enabled just opens the way to multiple patterns. As an example, the way I use it now is to deploy an ELB with TLS termination (cert from AWS) that points to nginx deployment.

Also using nginx-ingress-controller exposed behind a Service type=LoadBalancer in AWS makes it possible to use [external-dns](https://github.com/kubernetes-incubator/external-dns) with Route53 for automatically create DNS records when creating ingresses.

> In practical terms, I'd like to get Network Policies mature and included by default before this is considered. Cloud metadata needs to be locked down.

Very good point. This totally escaped me. ATM I use [kube2iam](https://github.com/jtblin/kube2iam) and containers can access metadata based on their IAM role.

> Can you explain the tags? I'm aware they're a "requirement" for --cloud-provider=aws, but I've yet to see a justification.

Without `KubernetesCluster` or `kubernetes.io/cluster/<cluster_name>: owned` the nodes where registering to the master and get deleted soon after because they we're not part of any cluster. 

I inspired from [kubernetes/kubernetes](ttps://github.com/kubernetes/kubernetes) the aws cloud provider [aws.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go) and [tags.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/tags.go) files.

Looking again today, I noticed another useful tag I think it should be added.

`// awsTagNameMasterRoles is a set of well-known AWS tag names that indicate the instance is a master
// The major consequence is that it is then not considered for AWS zone discovery for dynamic volume creation.
var awsTagNameMasterRoles = sets.NewString("kubernetes.io/role/master", "k8s.io/role/master")`

Without the following the ELB was getting created in the first subnet only.

`// TagNameSubnetPublicELB is the tag name used on a subnet to designate that
// it should be used for internet ELBs
const TagNameSubnetPublicELB = "kubernetes.io/role/elb"`

> Similar, for hostname setting requirement

The ec2 core os instances have only the short hostname without the <zone>.compute.internal domain by default. After setting `--cloud-provider=aws` the nodes were getting registered to the cluster and unregistered in a matter of seconds because they couldn't be found in the cloud. I guess the fqdn is the name of the node as returned and used by the kubernetes aws cloud provider. After I did this nodes register successfully to the cluster.

> Please don't add a new variable for setting the Container Linux image version. Picking the stable, beta, or alpha channel should be sufficient and keeps things simple (also prevents self-foot injuries)

I understand. I have a question though. If there is no version set, any time CoreOS does an update on stable/alpha/beta and releases the AMI the cluster would have to be recreated? In fact, the controller instance(s) and node asg.

aws/container-linux/kubernetes/ami.tf
```diff
@@ -14,6 +14,6 @@ data "aws_ami" "coreos" {
 
   filter {
     name   = "name"
-    values = ["CoreOS-${var.os_channel}-*"]
+    values = ["CoreOS-${var.os_channel}-${var.os_version}*"]
```
xiang90: this change has nothing to do with cloud provider, no?

valentinbud: no, it doesn't

acobaugh:
To @valentinbud 's comment from the 18th, external-dns can also function on Ingresses, but not every service is HTTP, and until Ingresses can expose raw TCP Services, Service with type=LoadBalancer is the only way to make this work with things that aren't http.

Agree that ec2 metadata should be locked down. Can we do that at the OS level, or is it better/preferred to use an egress NetworkPolicy? If we go the NetworkPolicy route, I wonder if that isn't better left to the cluster operator? Maybe create a policy for kube-system and leave the rest to the operator?

Either way, +1 for this PR minus the os_version variable. This change would let me easily switch off of kops.

dghubble:
> Service with type=LoadBalancer is the only way to make this work with things that aren't http

The Nginx Ingress controller supports TCP and UDP services if you provide a ConfigMap to the controller. I'm a user of the feature myself. The manifests shown in addons/ingress show how to deploy Ingress without needing to use service `type=LoadBalancer` anywhere.

valentinbud:
The only reason I have added `os_version` was because if I hadn't whenever the CoreOS AMI got updated running `terraform apply` recreated the cluster from scratch. I just wanted to have a stable cluster for a medium period of time while running terraform apply pretty often.

dghubble:
Thanks for the PR, but I'm not intending to add this.

PR #82, opened 2017-12-15
Enable BBR as default congestion algorithm

### LINKED ISSUE #67: Enable BBR congestion control algorithm
## Feature Request

I'd like to use the BBR congestion control algorithm on my Typhoon deployed cluster. This was enabled in CoreOS itself in https://github.com/coreos/bugs/issues/1719 .

jpds:
This change sets BBR as the default traffic congestion algorithm which has been in the kernel since 4.9.

Closes #67.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Notable changes between versions.
   * Older clusters (with CLUO addon) auto-update node's Container Linux version
   and will begin using Docker 17.09.
 * Fix race where `etcd-member.service` could fail to resolve peers ([#69](https://github.com/poseidon/typhoon/pull/69)) 
+* BBR is enabled as the default traffic congestion algorithm.
 
 #### Bare-Metal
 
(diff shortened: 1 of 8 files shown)


xiang90:
BBR is not always better than CUBIC. It depends on use cases. Enabling BBR by default might affect some workloads unintentionally.

dghubble:
Don't think anyone would argue either is always better. The question raised in the issue #67 is effectively just which makes a better default to ship enabled.

dghubble:
Since this is an additive change, it might be a good use for #140

dghubble:
#145 will allow users to provide custom Container Linux Config snippets. The feature will allow users to perform this type of customization, but in an elegant and validated way that merges Container Linux Config snippets into a single declarative Ignition document.

Adding files like `/etc/sysctl.d/bbr.conf`, systemd units, network configuration, raid arrays, or other host disk provisioning customizations will be possible and supersede the need for this issue.

PR #89, opened 2017-12-24
bm-docs: Add missing port to matchbox_http_endpoint

metral:
`matchbox_http_endpoint` does not specify the port number used in the BM TF example. This leads to machines not properly PXE booting using the example module provided, as they cannot reach the matchbox service.


dghubble:
Matchbox deployed on Kubernetes and exposed via Ingress is via port 80 and doesn't need to be specified (example came from my own infra and it's valid) . Matchbox on a standalone server does default to 8080 though.

I'd prefer this be clarified in docs since either setup is valid for different audiences.

PR #103, opened 2018-01-19
Update worker.yaml.tmpl to make iSCSI volumes work

yokhahn:
Adding iscsiadm and the iscsid config to be accessed from the kublet container
If iscsid has been configured on container linux as described here (https://coreos.com/os/docs/latest/iscsi.html) and the iscsi target is reachable from the worker this should allow iscsi volumes to work.

## Testing

In a local bare-metal typhoon deploy I was able to get iscsi volumes to work with this change made locally

### DIFF
--- a/bare-metal/container-linux/kubernetes/cl/worker.yaml.tmpl
+++ b/bare-metal/container-linux/kubernetes/cl/worker.yaml.tmpl
@@ -43,6 +43,10 @@ systemd:
           --mount volume=opt-cni-bin,target=/opt/cni/bin \
           --volume var-log,kind=host,source=/var/log \
           --mount volume=var-log,target=/var/log \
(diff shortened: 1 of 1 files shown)


dghubble:
I'd prefer to address this by moving away from the rkt-fly style of running the kubelet, on all platforms. That way, on-host binaries are implicitly available and it "just works", without the need to mount each set of binaries and files the Kubelet expects (but doesn't document).

This is a good way to get iSCSI working right now for anyone who needs it. :+1: Thanks

uwkhahn:
So, that will be great when that non rkt-fly method is available.   Is there a good place to understand the path to get there?

dghubble:
Rebased. updated, and merged as #350

PR #105, opened 2018-01-21
Update Digital Ocean variables / docs to use new droplet sizes

irontoby:
Digital Ocean have just [announced new droplet sizes](https://blog.digitalocean.com/new-droplet-plans/) with greater resources per price point compared with their previous offerings. They have also deprecated the existing sizes, which will be removed by July 1, 2018.

Users can now choose various combinations of CPU, memory, and disk resources, so the [droplet size names](https://developers.digitalocean.com/documentation/changelog/api-v2/new-size-slugs-for-droplet-plan-changes/) are a bit more complex.

* Update DO variable defaults to use the new sizes.
* Update docs with links to all available sizes.

I have just created a new cluster with these new default sizes and it seems to work fine. In fact it appears that even the slightly smaller `s-1vcpu-2gb` droplet can be used for a controller.

### DIFF
--- a/digital-ocean/container-linux/kubernetes/variables.tf
+++ b/digital-ocean/container-linux/kubernetes/variables.tf
@@ -27,8 +27,8 @@ variable "controller_count" {
 
 variable "controller_type" {
   type        = "string"
-  default     = "2gb"
-  description = "Digital Ocean droplet size (e.g. 2gb (min), 4gb, 8gb)."
+  default     = "s-2vcpu-2gb"
+  description = "Digital Ocean droplet size (e.g. s-2vcpu-2gb, s-2vcpu-4gb, s-4vcpu-8gb)."
 }
 
 variable "worker_count" {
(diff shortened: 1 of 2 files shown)


dghubble:
Thanks! I saw the announcement, but missed the new slug changes. The `s-1vcpu-2gb` should be equivalent to the old `2gb` controller default, instead of `s-2vcpu-2gb`.

Even so, this is great news for Digital Ocean users :100:

irontoby:
Well, I still have some of the "old" 2GB droplets and viewing them with `doctl compute droplet list` indicates they have "2 VCPUs", which is why I thought the `s-2vcpu-2gb` was equivalent. But I'm not sure their "VCPU" metric is the same across generations, and my small cluster seems to be doing fine with `s-1vcpu-2gb`, so that's probably a good pick for a bare bones setup.

That means we can now run a "starter" Digital Ocean cluster for $20/month!

Anyway thanks for all the great work on this project, and feel free to make whatever changes to the PR you see fit...

dghubble:
Actually, yeah, you're right. The old `2gb` Typhoon DO test cluster does show 2 vCPU as well.

Let's go with the `s-2vcpu-2gb` controller for now. I'll run the test cluster on `s-1vcpu-2gb` for a while and test some of the addons (can probably fit prometheus now!). If that turns out fine, we can switch before the next release - I know folks like to use the Digital Ocean clusters for development so $20/month would be great!

irontoby:
So it turns out `s-1vcpu-2gb` didn't hold up well under "real" loads after all. May work for a small dev cluster though...

Issue #94, opened 2018-01-11
VMware vSphere module?


johnmarcou:
Hello,

I am about to start to write a VMware vSphere module for Typhoon.

I just would like to know if you planned to support this platform, if you have already started this module, and if have any suggestions or recommendations?

Thank you.

dghubble:
I'm not planning for Typhoon to add or support a vSphere module in the near future.

Writing a module isn't the issue, its the maintenance, validation, testing, and support that goes into each platform, for Typhoon releases, for the foreseeable future. Typhoon clusters are run continuously on each platform, with workloads (simmer tests, not just "does it launch").  Platforms get independently tested and certified for conformance. Addons are verified to play nicely and provide a good admin experience. My bar is that I personally use and can recommend each aspect of Typhoon.

With VMWare, I'm not sure those goals can be achieved right now. VMWare vSphere is high-cost, which would make it difficult to offer the same quality guarantees and maintain it effectively. It requires ecosystem knowledge that should be maintained with involvement from VMWare. And its barrier to entry and lock-in mean a smaller audience is available to help in its upkeep (i.e. I don't know individuals running vSphere of their own choosing, its only within organizations). Those are just the tradeoffs, vSphere clearly has pros for many folks.

In my opinion, products are best maintained by those who both understand them deeply and love using them daily - Typhoon trying to offer vSphere right now, I think it'd miss the mark and the expectations of VMWare fans (Typhoon is non-commercial, not targeting every case, so I can be frank). VMWare seems to have its own Kubernetes offering (no idea if its good/bad), so it gets some love from VMWare engineers (or at least contractually obligated attention). Or you can build your own module and run it for a while, I do encourage it.

For now, you can try treating VMWare like bare-metal machines and use the bare-metal platform. Some folks have reported success - though I'll caution that's not how Typhoon gets validated.

johnmarcou:
Thanks a lot for your fast reply. This totally make sense.
I close the thread.

bkcsfi:
@johnmarcou Have you made any progress on getting Typhoon to work on vmware? I am also interested in working on this. I have a 3-node vsphere cluster that I'd like to try boot-kube on, but typhoon would be better, keeping in mind that it would not be supported, etc..

johnmarcou:
Hello. Yes, it's working now.

[...]

PR #110, opened 2018-01-24
Add Azure support

justaugustus:
Adds support for instantiating Typhoon clusters on Microsoft Azure.

### DIFF
--- a/azure/container-linux/kubernetes/controllers.tf
+++ b/azure/container-linux/kubernetes/controllers.tf
@@ -90,7 +90,7 @@ resource "azurerm_network_interface" "controller" {
 
   ip_configuration {
     name                                    = "controllerIPConfig"
-    subnet_id                               = "${azurerm_subnet.public.id}"
+    subnet_id                               = "${azurerm_subnet.controller.id}"
     private_ip_address_allocation           = "dynamic"
     public_ip_address_id                    = "${element(azurerm_public_ip.controller.*.id, count.index)}"
     load_balancer_backend_address_pools_ids = ["${azurerm_lb_backend_address_pool.apiserver.id}"]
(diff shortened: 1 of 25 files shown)


justaugustus:
@dghubble functionally, everything works, but I still need to update docs for this.
I figured I'd toss it up here so you could take a peep!

azure/container-linux/kubernetes/controllers.tf
```diff
@@ -0,0 +1,330 @@
+# TODO: Add Scale Sets implementation once support exists: https://github.com/kubernetes/kubernetes/issues/43287
+
+# Discrete DNS records for each controller's private IPv4 for etcd usage
+resource "azurerm_dns_a_record" "etcds" {
```
colemickens: Why are separate etcd records needed? (aka, why not just use static IPs?)

docs/azure.md
```diff
@@ -0,0 +1,268 @@
+# AWS
```
colemickens: `s/AWS/Azure/g`

docs/azure.md
```diff
@@ -0,0 +1,268 @@
...
+Reference the DNS zone id with `"${aws_route53_zone.zone-for-clusters.zone_id}"`.
+
+!!! tip ""
+    If you have an existing domain name with a zone file elsewhere, just carve out a subdomain that can be managed on Route53 (e.g. aws.mydomain.com) and [update nameservers](http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/SOA-NSrecords.html).
+
+### Optional
+
+| Name | Description | Default | Example |
+|:-----|:------------|:--------|:--------|
+| controller_count | Number of controllers (i.e. masters) | 1 | 1 |
```
colemickens: More Amazon stuff in here, plus I think you require an extra thing or two for Azure.

colemickens: Actually there's quite a bit above two in the examples, even in places other than example outputs.

colemickens:
Looks like it requires a full DNS zone to donate to the cluster? Or it just has to be present so that this can poke records into it? (Is it safe to assume the records have the cluster_name in it, or something to preserve uniqueness?)

docs/azure.md
justaugustus: @colemickens Ya, still working the docs. This is just copy pasta for now.

dghubble:
This gets the same answer [VMWare](https://github.com/poseidon/typhoon/issues/94) and OVH did.

Writing a module isn't the issue, its the maintenance, validation, testing, and support that goes into each platform, for Typhoon releases, for the foreseeable future. Typhoon clusters are run continuously on each platform, with workloads (simmer tests, not just "does it launch"). Platforms get independently tested and certified for conformance. Addons are verified to play nicely and provide a good admin experience. 

A priority ranked roadmap does include new platforms (further out), with QEMU/KVM support as the next platform planned, after other priorities. Azure is a good contender for another platform in my mind, but at a minimum we'd need to get Azure non-profit sponsorship to fund clusters. And preferably get a solid, trusted point of contact there.

We can revisit this in the future.

azure/container-linux/kubernetes/controllers.tf
justaugustus: @colemickens These etcd records are used as SANs in the etcd TLS asset generation process.

justaugustus:
@colemickens I've updated the docs for Azure, if you'd like to take a look.
w.r.t. DNS records, you can delegate a domain or subdomain for usage in Azure. 
We'd simply be putting records into a pre-delegated zone.

While `cluster_name` is used to generate the records, it's not truly possible to ensure uniqueness if someone decides to create another cluster in that zone with the same name.

justaugustus:
(for anyone driving by this PR...)
@dghubble and I have discussed this offline and there's some legwork to do before this might be considered for use in Typhoon, so I'm going to close this out.

In the meantime, if you're itching to use Typhoon on Azure, you can check out [typhoon-azure](https://github.com/justaugustus/typhoon-azure)!

Issue #108, opened 2018-01-24
Allow TLS materials injection in Typhoon


johnmarcou:
### Feature

Would it be possible to update the Typhoon (and on-prem) repo to accept/reflect all the `bootkube` variables please? I am talking about these [variables](https://github.com/poseidon/terraform-render-bootkube/blob/master/variables.tf) to these [ones](https://github.com/poseidon/typhoon/blob/master/bare-metal/container-linux/kubernetes/variables.tf).

### Tradeoffs

My understanding is Typhoon = Booktube + Matchbox + CL templates.
It would be nice to be able to customise the Kubernetes deployment like when we use bootkube manually.

Some are already accepted like `pod_cidr`, `service_cidr ` and  `networking `.
In my current context, I would like to force the `ca_certificate `, `ca_key_alg ` and `ca_private_key` to keep my `kubeconfig` files static, even after update/clusters re-creation.

Thank you.

dghubble:
As a distribution, Typhoon makes choices about the options to expose in each module. `terraform-render-bootkube` is a lower-level component not targeted at direct usage, so it has more options. Its intentional that `pod_cidr`, `service_cidr`, and `networking` are exposed, but others are not (like `etcd_servers`).

Re-using TLS materials from old clusters isn't a supported use case today. I understand its easy because you don't have to change your `kubeconfig`, but it also creates an environment where you'll forget to rotate (they expire), rely on the materials not changing, or be disappointed if there are future changes for the TLS material inputs. I'm hesitant to support anything other than the auto-generated options.

Check the upgrade/reprovision docs for the current recommendations: https://typhoon.psdn.io/topics/maintenance/#upgrades

dghubble:
I suppose, I'd like to better understand the desire to force old TLS materials. I'd like to think about whether there are better ways to address the underlying challenge with tooling or other solutions.

johnmarcou:
Thanks for you fast reply, as usual, and detailed answer!
I just would like to bring some arguments to this question.

I agree it's doesn't make sense to expose all the `bootkube` argument on `Typhoon`, `etcd_servers` is a good example of internal variable.

However, I think if there is enough reasons to allow "non-generated" TLS materials on bootkube, same reasons can apply to Typhoon.

Another example, if the administrator relay on an external TLS rotation system, it might be interesting for him to be able to use it at the Typhoon deployment. 

[...]

PR #111, opened 2018-01-25
Add CA configuration for DigitalOcean

mochja:
Add bootkube-renderer supported CA variables into DigitalOcean module. This allows to specify custom CA certificate used by Kubernetes instead of generating one on the fly.

## Testing

I have validated cluster after creation by `kubectl get nodes` which resulted in list of nodes. CA crt in assets directory matches one I have provided via terraform variable.

### DIFF
--- a/digital-ocean/container-linux/kubernetes/bootkube.tf
+++ b/digital-ocean/container-linux/kubernetes/bootkube.tf
@@ -11,4 +11,7 @@ module "bootkube" {
   pod_cidr              = "${var.pod_cidr}"
   service_cidr          = "${var.service_cidr}"
   cluster_domain_suffix = "${var.cluster_domain_suffix}"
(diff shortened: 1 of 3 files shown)


dghubble:
rel #108

dghubble:
The crux of #108 is justifying the use-case. Can you explain a bit more?

mochja:
In short, we would like to manage client access to the cluster in centralized way eg. [Vault](https://github.com/hashicorp/vault) so we can have on-demand access into Kubernetes.

With intermediate certificates and multiple clusters we could also achieve level-based access. Theoretically speaking you could have single "master" certificate to access each and every of your clusters, but still have power to generate client certs only for specific cluster.

dghubble:
May revisit this in future, but I still don't want to support these inputs across bare-metal and clouds.

dghubble:
Thanks though!

PR #126, opened 2018-02-14
*: Add `ignore_changes` to lifecycle block for OS images

justaugustus:
Configures TF lifecycle block to ignore changes to the OS image on each platform.
Should fix the problem referenced here: https://github.com/poseidon/typhoon/pull/76#issuecomment-365525232

This wasn't extensively tested, outside of the fact that I've used the lifecycle block and `ignore_changes` before in other projects.

@dghubble let me know if this looks sane to you.

### DIFF
--- a/aws/container-linux/kubernetes/controllers.tf
+++ b/aws/container-linux/kubernetes/controllers.tf
@@ -36,6 +36,10 @@ resource "aws_instance" "controllers" {
   associate_public_ip_address = true
   subnet_id                   = "${element(aws_subnet.public.*.id, count.index)}"
   vpc_security_group_ids      = ["${aws_security_group.controller.id}"]
+
+  lifecycle {
+    ignore_changes = ["ami"]
+  }
 }
 
 # Controller Container Linux Config
--- a/digital-ocean/container-linux/kubernetes/controllers.tf
+++ b/digital-ocean/container-linux/kubernetes/controllers.tf
@@ -50,6 +50,10 @@ resource "digitalocean_droplet" "controllers" {
   tags = [
     "${digitalocean_tag.controllers.id}",
   ]
+
+  lifecycle {
+    ignore_changes = ["image"]
+  }
 }
 
 # Tag to label controllers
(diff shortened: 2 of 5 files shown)


dghubble:
I don't think this is needed, except perhaps AWS. I don't want to add bits that aren't fully understood.

### AWS

I suspect it is needed on AWS because the AMI is always dynamically resolved in Terraform. A user chooses a channel (such as the default "stable"), deploys a cluster, retains the cluster until a new CL release occurs, and re-runs `terraform apply`. At that point Terraform re-evaluates the mapping from channel to actual AMI id, detects a change, and decides the instances should be replaced. This is undesired since controller nodes should not be replaced, plus node upgrades can already be handled by the CLUO addon.

```
"ami-65269d1d" => "ami-fc01ba84"
```

I've left an AWS cluster running with CL alpha to be able to verify this buggy behavior and test just the AMI fix.

### Others

On Digital Ocean, images like "coreos-stable" are created via the DO API (no dynamic resolution) so I've yet to see a case where DO will destroy existing nodes upon new images becoming available in a channel.

On Google Cloud, docs currently recommend choosing exact images (`gcloud compute images list | grep coreos`) which avoids any issues. Its possible to specify "coreos-stable" to the GCP API for use in the controller image and worker instance template. I'm testing this out. I've yet to see GCP destroy existing nodes upon new images becoming available in a channel, but will report if its safe to switch from exact images to "coreos-stable", etc.

dghubble:
There is indeed a bug on AWS (regression, since old controller ASGs had the `ignore_changes`). The AMI id changes since the channel is evaluated client-side. I've taken just the AWS controller fix in c8313751d7ec4af5766cc93c7cd29ccbc2cb7a7d.

I'm not able to demonstrate this fixes the issue because going from a hash without the `ignore_change` to a hash with the `ignore_changes` after the AMI change still shows a diff that causes controller replacement. I suspect `ignore_changes` has to be initially present. I've deployed a fresh test cluster to catch the next alpha release cycle to verify.

dghubble:
Merged c8313751d7ec4af5766cc93c7cd29ccbc2cb7a7d

Issue #130, opened 2018-02-15
Explore using an NLB instead of an ELB


dghubble:
NLBs are the newer network load balancer option on AWS. Support has been maturing in newer version of the terraform-provider-aws provider and we may be able to start exploring switching the ELBs to NLBs.

Potential benefits:

* AWS claims NLBs have better performance and throughput.

There are a few things to look out for:

* Just as reliable in practice?
* No increase in time to provision an AWS cluster
* Would require bumping the minimum terraform-provider-aws plugin verison we allow

We can investigate this separately for the apiserver and ingress ELBs.

dghubble:
As of 461fd469862a0496031b967fcdf5e2e6baf0ef32, the apiserver and ingress ELBs on AWS clusters have been swapped for NLBs. This should be reflected in the next release after v1.9.3.

PR #134, opened 2018-02-16
AWS NLB - ingress

pms1969:
Ingress load balancer set to an AWS NLB

## Testing

deployed/destroyed/applied all addons.

ref: #130


dghubble:
Thanks, but I have a branch from November I've been testing with and I'm probably going to use that one.

dghubble:
This doesn't work in practice. The target group will fail health checks. Additionally, Ingress should load balance both 80 and 443 traffic so that HTTP to HTTPS redirects work properly for apps exposed via Ingress.

Going with #141

PR #135, opened 2018-02-16
AWS - Private subnets

pms1969:
Launches controllers and workers in a private subnet.

* Creates a NAT and Bastion
* Launches load balancers, nat, gateway and bastion in public subnets
* Launces controllers and workers in private subnets.
* Uses the AWS KeyPair association for instances, removing the need for ssh-authorized-key, but adding the need for the pem to be on the working machine.
* Tunnels to controllers and workers via the bastion.

## Testing

I've created a cluster and deployed all the addons to it, with no issues.

Time to create has increased a touch.

### DIFF
--- a/aws/container-linux/kubernetes/bastion.tf
+++ b/aws/container-linux/kubernetes/bastion.tf
@@ -1,0 +1,122 @@
+# kube-apiserver Network Load Balancer DNS Record
+resource "aws_route53_record" "bastion" {
+  zone_id    = "${var.dns_zone_id}"
+
+  name = "${format("bastion.%s.%s.", var.cluster_name, var.dns_zone)}"
+  type = "A"
+
+  # AWS recommends their special "alias" records for ELBs
+  alias {
+    name                   = "${aws_lb.bastion.dns_name}"
+    zone_id                = "${aws_lb.bastion.zone_id}"
+    evaluate_target_health = true
+  }
+}
+
+# Controller Network Load Balancer
+resource "aws_lb" "bastion" {
+  name               = "${var.cluster_name}-bastion"
+  load_balancer_type = "network"
+  subnets            = ["${aws_subnet.public.*.id}"]
+}
+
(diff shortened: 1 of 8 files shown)


dghubble:
I don't have an intention of merging this change. I've been down this road.

Yes, it somewhat mirrors architectural choices I made and implemented for Tectonic AWS in the early days, but I don't want it in my distribution. Enterprises have a desire for workers to be in private subnets, but its mostly security theater:

You get to carve out special private subnets just for workers, deal with AWS's NAT gateway and routing table concept, and run a dedicated EC2 instance and load balancer just for the bastion (at least in the Tectonic approach the controllers were used as the bastion). Meanwhile, the bastion t2.micro is not managed as part of the cluster, isn't part of coordinated Container Linux upgrades, and you have to explain to end-users how to SSH through a bastion when there is a problem (and they're panic'ing). And the entire cluster is unavailable when the bastion is upgrading. It adds to the cost and complexity of an AWS cluster. If this were merged, issues would be filed asking for the ability to make it optional or the ability to customize the bastion's properties ("t2.micro" won't be enough for some, etc).

* If the worry is sshd vulnerabilities, breaking into the bastion sinks the victim just as much as breaking into the worker (plus bastion is less likely to be updated). I'll point out Container Linux only allows key-based logins as well, never passwords, in case that was a concern.
* If the worry is vulnerabilities in apps exposed over ingress (worker firewalls only allow SSH and HTTP/HTTPS for ingress), the private subnet isn't going to help. You're already exposing pods so you're trusting they don't have exploitable vulnerabilities.

I'd recommend evaluating the attack vectors you're seeking to protect against.

aside: The AWS key pair change is unrelated and also undesired. The Container Linux config approach is taken because its general and cross-platform, no reason to rely on the AWS agent for this.

pms1969:
Fair enough...  I have a fairly enterprise use case, so this just ticks a few boxes for clients.  Points taken, and I'll do some further investigation anyway.

PR #137, opened 2018-02-16
Fix: namespaces

pms1969:
Renames namespace configs so that they are picked up first.

    This prevents kubectl from applying them in the wrong order
    when recursively applying a directory.

    i.e. kubectl apply -f addons/prometheus/ -R
    would fail with namespace not available errors.  This should not happen now.

## Testing

deployed all of the addons successfully.


pms1969:
damn, somehow got it in the wrong place.

PR #138, opened 2018-02-16
Fix: namespaces

pms1969:
Renames namespace configs so that they are picked up first.

    This prevents kubectl from applying them in the wrong order
    when recursively applying a directory.

    i.e. `kubectl apply -f addons/prometheus/ -R`
    would fail with namespace not available errors.  This should not happen now.

## Testing

deployed all of the addons successfully.


addons/prometheus/_namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
dghubble: There is already a namespace in the prometheus directory, with your change there would be two.

pms1969: yeah, spotted that.  I think there might actually be a case for moving the namespaces out into their own addon?

pms1969:
Will make changes soon, and push changed commit message.

addons/prometheus/_namespace.yaml
dghubble: Doesn't make sense to me that they'd be in a separate addon. Rationale?

dghubble:
Hm? Please squash your commits into one commit anyway.

pms1969:
Done.

dghubble:
Looks good. What about the namespace in nginx-ingress for google-cloud and digital-ocean?

pms1969:
Done.  Sorry, been so focused on getting other stuff working, I hadn't even clocked that those directories existed.

dghubble:
Thanks! Rebased and merged as 86420fd5072799a73b1dd729d5f6b188d71b672f

PR #149, opened 2018-02-27
ease of use for the first timer

lvnilesh:

* Specific change
* Specific change

## Testing


dghubble:
This being easier is down to opinion. I think a cluster deployer can be expected to know how to get pods without needing a long command cluttering my docs. These sed one liners also encourage a copy-paste without thinking mentality.

Also, `open` isn't cross platform and there is a link right below that will open your browser.

PR #155, opened 2018-03-05
docs: use route53 example for ingress/aws

bendrucker:
Even though there are sections for each cloud provider, I noticed the examples reference Google Cloud. I could update the DO section as well, just not familiar with that TF provider off hand.


dghubble:
Hm, its not a mistake they reference GCP. You may choose to use any DNS provider for application DNS records, regardless of the cloud the cluster is deployed onto. I'd have gone with an unaffiliated DNS provider (say dnssimple) to illustrate this more clearly, but figured it'd be unfamiliar. So of the bunch, I slightly favored showing GCP.

bendrucker:
Gotcha, thought that might be the case. I think it’s very likely that someone will use their cloud provider’s dns even though they could pick a generic service.

PR #156, opened 2018-03-07
Add variables for mounting additional volumes to kubelet pod

jjlakis:
Added new variables that allow to add additional mounting points to kubelet pods. It is necessary if we want kubelet to map Ceph RBD volumes (ceph configuration directory needs to be accessible from kubelet pod).

`controller-configs` template is extended with another parameter that is created from the mount list. It adds `--volume` and `--mount` parameters to `rkt run` args and `ExecStartPre`s with directory creation. Value is optional and empty by default.

Unfortunately, I had to create 2 variables for each type of kubelet pod. `for` is not implemented in terraform yet and parameters have to be created by `formatlist`, where I can't find the current list index during iteration, can't even use `range` list so I am unable to create unique volume name and need to pass it from another list.
Same issue is described here:
https://github.com/hashicorp/terraform/issues/2769
and last comment announces solution for this problems in near future.

Tested & working without any changes, on 3-node configuration that works with master code

To add `ceph-conf` volume to controller node with `/etc/ceph:/etc/ceph` mapping add following variables:
```
kubelet_controller_mount_points = ["ceph-conf"]
kubelet_controller_mount_points_mapping = { "/etc/ceph" = "/etc/ceph" }
```

### DIFF
--- a/bare-metal/container-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/bare-metal/container-linux/kubernetes/cl/controller.yaml.tmpl
@@ -66,7 +66,7 @@ systemd:
           --volume opt-cni-bin,kind=host,source=/opt/cni/bin \
           --mount volume=opt-cni-bin,target=/opt/cni/bin \
           --volume var-log,kind=host,source=/var/log \
-          --mount volume=var-log,target=/var/log \
+          --mount volume=var-log,target=/var/log \${controller_mount_parameters}
           --insecure-options=image"
         ExecStartPre=/bin/mkdir -p /opt/cni/bin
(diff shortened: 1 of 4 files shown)


dghubble:
I'm not really willing to take these new variables (clutter, maintenance burden, lower-readability, unclear usage). Supporting every option isn't a goal. Tweaking in a fork for your needs is fairly straight-forward.

I'll point out, flex volume plugins aren't used by everyone, but its safe, unobtrusive, and simple to mount in all cases. I'd rather consider that approach for Ceph RBD fans.

dghubble:
As in, what about just adding the mount ceph needs?

PR #162, opened 2018-03-12
Update Kubernetes from v1.9.3 to v1.9.4

dghubble:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md#v194

### Notice

Upstream made a breaking change https://github.com/kubernetes/kubernetes/pull/58720. It impacts the Grafana addon.

> "Secret, configMap, downwardAPI and projected volumes will be mounted as read-only volumes. Applications that attempt to write to these volumes will receive read-only filesystem errors. Previously, applications were allowed to make changes to these volumes, but those changes were reverted at an arbitrary interval by the system. Applications should be re-configured to write derived files to another location."

Grafana attempts to chown any dashboards its given, so I'm not sure of the workaround.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,9 +4,10 @@ Notable changes between versions.
 
 ## Latest
 
+* Kubernetes [v1.9.4](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md#v194)
 * Introduce [worker pools](https://typhoon.psdn.io/advanced/worker-pools/) for AWS and Google Cloud. Allow groups of workers with different properties to be joined to a cluster.
 * Use new Network Load Balancers and cross zone load balancing on AWS
-* Allow flexvolume plugins to be used on all Typhoon clusters (not just bare-metal)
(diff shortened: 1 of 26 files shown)


dghubble:
Note that upstream made a breaking change https://github.com/kubernetes/kubernetes/pull/58720

dghubble:
I'm unhappy upstream shipped https://github.com/kubernetes/kubernetes/pull/58720, but introducing feature gates at this late stage is even worse. Given the choice between vanilla v1.9.4 (broken Grafana) and v1.9.4 with a feature gate to rip out the breakage (but undo security fix?) I'm going with vanilla v1.9.4.

However, I've checked that you can still deploy Grafana v4.6.3. Checkout Typhoon v1.9.3 and Grafana works just fine on v1.9.4 clusters.

dghubble:
https://github.com/grafana/grafana-docker/issues/140

dghubble:
Another known issue is that `subPath` seems to have regressed. https://github.com/kubernetes/kubernetes/issues/61076

PR #166, opened 2018-03-23
kubelet.service: disable read-only port

schu:
The read-only port doesn't have any authorization, even when kubelet
authorization is enabled. Disable it to avoid unauthorized access to
private information.

In k8s 1.10 this will be the default:

https://github.com/kubernetes/kubernetes/pull/59666/commits/c1e34bc725010cfab2157545d4c0f8bc851db966

### DIFF
--- a/aws/container-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/aws/container-linux/kubernetes/cl/controller.yaml.tmpl
@@ -84,6 +84,7 @@ systemd:
           --node-labels=node-role.kubernetes.io/controller="true" \
           --pod-manifest-path=/etc/kubernetes/manifests \
           --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
+          --read-only-port 0 \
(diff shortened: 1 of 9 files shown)


schu:
Noticed after reading https://medium.com/handy-tech/analysis-of-a-kubernetes-hack-backdooring-through-kubelet-823be5c3d67c

dghubble:
Firewall rules prevent access to these portd by default on each cluster. However, read-only port is used for reliable Kublet health checking. If you try out your PR, I think you'll find it breaks all clusters.

schu:
> However, read-only port is used for reliable Kublet health checking.

Ah, I didn't see it's used by Prometheus and indeed haven't checked the addon.

dghubble:
I meant the Kubelet port is used by the apiserver load balancers as a rough
estimate of whether to include a controller as a backend. It's not ideal,
but various distros have used this in lieu of an appropriate apiserver
endpoint for a while.

I'm afk, but can investigate the new situation in 1.10 to see if we can
make any improvements.

And yeah it's also used for in-cluster metrics.

PR #170, opened 2018-03-26
Allow users to specify a custom SSH port

dmyerscough:

Allow users to specify what port SSH listen on and configure connection idle timeouts for SSH.

## Testing

```bash
$ terraform plan
$ terraform apply
```

### DIFF
--- a/digital-ocean/container-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/digital-ocean/container-linux/kubernetes/cl/controller.yaml.tmpl
@@ -115,6 +115,13 @@ systemd:
         ExecStartPost=/bin/touch /opt/bootkube/init_bootkube.done
         [Install]
         WantedBy=multi-user.target
+    - name: sshd.socket
+      dropins:
+      - name: 10-sshd-port.conf
+        contents: |
+          [Socket]
+          ListenStream=
+          ListenStream=${ssh_port}
 storage:
   files:
     - path: /etc/kubernetes/kubelet.env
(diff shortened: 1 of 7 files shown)


dghubble:
If you need to add dropins and overwrite `sshd_config, you can now use Container Linux Config snippets https://typhoon.psdn.io/advanced/customization/#hosts. If you need to add your personal preferred sshd_config, that's the place. I'd like to retain the default OS `sshd_config`,which already prevents root and password logins and gets handled by Container Linux.

If you need to change SSH port, its probably best to just maintain a fork for yourself. Doesn't feel like the benefits of an obfuscated port warrant a new variable and supporting it across platforms.

dmyerscough:
That's fine, Ill just maintain this in a separate fork. The default SSH config doesn't match the match the CIS Benchmark standards.

> If you need to change SSH port, its probably best to just maintain a fork for yourself. Doesn't feel like the benefits of an obfuscated port warrant a new variable and supporting it across platforms.

The benefits are avoiding constant scanning/brute force attempts since port 22 is constantly being scanned by botnets.

[1] https://www.cisecurity.org/cis-benchmarks/

dghubble:
Thanks. I'd love to see Container Linux give consideration for meeting those requirements by default. And no disagreement that there are benefits to obfuscating.

dghubble:
If your sshd_config came from some sorta official mandate, maybe file on coreos/bugs?

dmyerscough:
I'll look at filing a bug with Container Linux.

PR #175, opened 2018-03-29
Add etcd metrics, Prometheus scrapes, and Grafana dash

### LINKED ISSUE #114: Scape etcd targets in Prometheus addon
## Feature

Configure the `prometheus` addon to scape Typhoon etcd targets on controller nodes. Then, metrics from etcd will be available in Prometheus. Alert rules for etcd will fire during incidents. The etcd dashboard provided with the `grafana` addon will be populated.

Invariants:
* Users still only need to choose to `kubectl apply` the addon manifests. Nothing more.
* Users never need to fiddle with listing etcd nodes on any platform.

### Background

The `prometheus` addon manifests setup Prometheus 2.1 (#113) to scape apiservers, kubelets, services, endpoints, cAdvisor, and exporters (kube-state-metrics and node_exporter). Alerting rules and Grafana graphs in addons correspond to these metrics. However, etcd rules and graphs currently aren't active/populated.

### Situation

Prometheus's can be configured (via the ConfigMap) to scrape the secured `:2379/metrics` endpoints of etcd nodes just like any other target. The etcd cluster runs on-host, across controllers with systemd, it is a lower-level component on which Kubernetes relies (not atop k8s), and it handles its own client authentication already.

* Typhoon runs etcd on-host, across controllers, on all platforms
* Typhoon requires etcd be setup with TLS on all platforms
* Typhoon creates etcd client certs, but only places them on controller nodes

To perform the scrapes, Prometheus needs the etcd client certificates to write the `tls_config` [section](https://github.com/poseidon/typhoon/blob/master/addons/prometheus/config.yaml#L39) in a new scape job.

### Options

* Add etcd client materials in a `kube-system` secret. We did [this](https://github.com/poseidon/terraform-render-bootkube/blob/533e82f833c166297abd249ac3d4853d6ebed364/resources/experimental/manifests/etcd-client-tls.yaml) back when self-hosted etcd was explored.
  * Pro: Allows prometheus pod to be scheduled on any node
  * Con: Opens up the possibility of escalation attacks (i.e. read kube-system secrets == read everything)
* Mount the etcd client materials from a controller host. (current most viable)
  * Pro: Avoid keeping etcd client materials in a Kubernetes secret
  * Con: Restricts prometheus pod itself to run on controller nodes
* Explore whether its possible to create (or invent) "metrics-only" etcd certificates
  * Likely not on the roadmap
* Metrics whitelist proxy
  * Con: I use some whitelist proxies for some internal things. They're gross though.

dghubble:
* Use etcd v3.3 `--listen-metrics-urls` to expose only metrics data via http://0.0.0.0:2381 on controllers
* Add Prometheus discovery for etcd peers on controller nodes
* Enables etcd related alerts and populates the etcd Grafana dashboard
* Note: These benefits require the optional Prometheus and Grafana addons be applied.

Hold off on allowing workers firewall access (can't think of any concrete concern with workloads seeing this). Move Prometheus to a controller node for a while (maybe drop). Adjust firewall rules now that Prometheus can run on a controller, rather than a worker

Made possible by:

* https://github.com/coreos/etcd/issues/8060 thanks @gyuho @brancz
* https://github.com/coreos/etcd/pull/8242

screenshot from 2018-03-28 22-30-52

Closes #114

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,16 @@ Notable changes between versions.
 
 ## Latest
 
+* Enable etcd v3.3 metrics endpoint ([#175](https://github.com/poseidon/typhoon/pull/175))
+
+#### Addons
+
+* Add Prometheus discovery for etcd peers on controller nodes ([#175](https://github.com/poseidon/typhoon/pull/175))
+  * Scrape etcd v3.3 `--listen-metrics-urls` for metrics
+  * Enable etcd alerts and populate the etcd Grafana dashboard
+
+## v1.10.0
+
 * Kubernetes [v1.10.0](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.10.md#v1100)
 * Remove unused, unmaintained `pxe-worker` internal module
 
--- a/addons/prometheus/config.yaml
+++ b/addons/prometheus/config.yaml
@@ -112,6 +112,22 @@ data:
         target_label: __metrics_path__
         replacement: /api/v1/nodes/${1}/proxy/metrics/cadvisor
     
+    # Scrap etcd metrics from controllers 
+    - job_name: 'etcd'
+      kubernetes_sd_configs:
+      - role: node
+      scheme: http
+      relabel_configs:
+        - source_labels: [__meta_kubernetes_node_label_node_role_kubernetes_io_controller]
+          action: keep
+          regex: 'true'
+        - action: labelmap
+          regex: __meta_kubernetes_node_label_(.+)
+        - source_labels: [__meta_kubernetes_node_name]
+          action: replace
+          target_label: __address__
+          replacement: '${1}:2381'
+    
     # Scrape config for service endpoints.
     #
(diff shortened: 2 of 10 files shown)


dghubble:
The `HighNumberOfFailedGRPCRequests` Prometheus rule triggers noisy alerts for `etcdserverpb.Watch` with `grpc_code` Unavailable. The rule gets triggered across platforms, even with very fast SSDs and I can manually use `etcdctl` to watch keys without issue.

brancz:
We've seen that a couple times before, and I believe that it's ultimately because of https://github.com/coreos/etcd/issues/9166 which doesn't seem like it landed in 3.3

dghubble:
For now I think I'll drop the two noisy alerts. Even without them, this change adds etcd alerts which weren't active before and populates the one page in Grafana that was empty before.

Its a good feedback loop too! - using the new etcd alerts motivated switching to faster disks on AWS for the v1.10 release (other platforms were fast enough).

PR #182, opened 2018-04-10
Update example BGPPeer manifest

madorn:
Previous example may have been outdated. It resulted in `error: unable to recognize "example.yaml": no matches for /, Kind=bgpPeer` .

See https://docs.projectcalico.org/v3.0/reference/calicoctl/resources/bgppeer.


dghubble:
Thanks! That example was indeed from Calico v2.x series.

PR #193, opened 2018-04-17
Fixed images for addons documentation

lukemorbin:
Fixed images for addons documentation.

Changed path from / -> ../ which is correct to display the images inline in the markdown.


dghubble:
Thanks! Squash merged in 490b628e2d318fbc26076965037ffb34d685b3a0

PR #197, opened 2018-04-21
Update Calico from v3.0.4 to v3.1.1

dghubble:
* https://github.com/projectcalico/calico/releases/tag/v3.1.1
* https://github.com/projectcalico/calico/releases/tag/v3.1.0
* https://github.com/poseidon/terraform-render-bootkube/pull/56

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ Notable changes between versions.
 
 ## Latest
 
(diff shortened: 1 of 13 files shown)


dghubble:
I'm kinda on the fence about Calico reading `/var/lib/calico/nodename` now. They've added it to address users misconfiguring the node name by hand between `calico/node` and the CNI config and to facilitate migrations. Typhoon sets up both correctly and doesn't permit a customization that would break this - the new mount doesn't really do anything useful for us.

PR #209, opened 2018-05-08
Allow for using Flatcar Linux instead of Container Linux

dongsupark:
Make the bare-metal provider work based on not only Container Linux but also [Flatcar Linux](https://flatcar-linux.org). To use Flatcar Linux, user needs to set a config variable `os_channel` to one of the following variables: `flatcar-stable`, `flatcar-beta`, `flatcar-alpha`.

Also create a new profile `fl` for Flatcar Linux, and rename a corresponding config variable `container_linux_version` to `os_version`.

### DIFF
--- a/bare-metal/container-linux/kubernetes/fl/controller.yaml.tmpl
+++ b/bare-metal/container-linux/kubernetes/fl/controller.yaml.tmpl
@@ -1,0 +1,171 @@
+---
+systemd:
+  units:
+    - name: etcd-member.service
+      enable: true
+      dropins:
+        - name: 40-etcd-cluster.conf
+          contents: |
+            [Service]
+            Environment="ETCD_IMAGE_TAG=v3.3.4"
+            Environment="ETCD_NAME=${etcd_name}"
+            Environment="ETCD_ADVERTISE_CLIENT_URLS=https://${domain_name}:2379"
+            Environment="ETCD_INITIAL_ADVERTISE_PEER_URLS=https://${domain_name}:2380"
+            Environment="ETCD_LISTEN_CLIENT_URLS=https://0.0.0.0:2379"
+            Environment="ETCD_LISTEN_PEER_URLS=https://0.0.0.0:2380"
+            Environment="ETCD_LISTEN_METRICS_URLS=http://0.0.0.0:2381"
+            Environment="ETCD_INITIAL_CLUSTER=${etcd_initial_cluster}"
+            Environment="ETCD_STRICT_RECONFIG_CHECK=true"
+            Environment="ETCD_SSL_DIR=/etc/ssl/etcd"
+            Environment="ETCD_TRUSTED_CA_FILE=/etc/ssl/certs/etcd/server-ca.crt"
+            Environment="ETCD_CERT_FILE=/etc/ssl/certs/etcd/server.crt"
+            Environment="ETCD_KEY_FILE=/etc/ssl/certs/etcd/server.key"
+            Environment="ETCD_CLIENT_CERT_AUTH=true"
+            Environment="ETCD_PEER_TRUSTED_CA_FILE=/etc/ssl/certs/etcd/peer-ca.crt"
+            Environment="ETCD_PEER_CERT_FILE=/etc/ssl/certs/etcd/peer.crt"
+            Environment="ETCD_PEER_KEY_FILE=/etc/ssl/certs/etcd/peer.key"
+            Environment="ETCD_PEER_CLIENT_CERT_AUTH=true"
+    - name: docker.service
+      enable: true
+    - name: locksmithd.service
+      mask: true
+    - name: kubelet.path
+      enable: true
+      contents: |
+        [Unit]
+        Description=Watch for kubeconfig
+        [Path]
+        PathExists=/etc/kubernetes/kubeconfig
+        [Install]
+        WantedBy=multi-user.target
+    - name: wait-for-dns.service
+      enable: true
+      contents: |
+        [Unit]
+        Description=Wait for DNS entries
+        Wants=systemd-resolved.service
+        Before=kubelet.service
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/bin/sh -c 'while ! /usr/bin/grep '^[^#[:space:]]' /etc/resolv.conf > /dev/null; do sleep 1; done'
+        [Install]
+        RequiredBy=kubelet.service
+        RequiredBy=etcd-member.service
+    - name: kubelet.service
+      contents: |
+        [Unit]
+        Description=Kubelet via Hyperkube
+        Wants=rpc-statd.service
+        [Service]
+        EnvironmentFile=/etc/kubernetes/kubelet.env
+        Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \
+          --volume=resolv,kind=host,source=/etc/resolv.conf \
+          --mount volume=resolv,target=/etc/resolv.conf \
+          --volume var-lib-cni,kind=host,source=/var/lib/cni \
+          --mount volume=var-lib-cni,target=/var/lib/cni \
+          --volume var-lib-calico,kind=host,source=/var/lib/calico \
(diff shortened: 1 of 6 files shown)


dghubble:
Thanks! I've been thinking about some different ways of allowing Flatcar Linux (and potentially other Container Linux derivatives) around v1.11.x timeframe. Bare-metal is certainly the biggest commitment, so I'd like to explore allowing it on clouds first ideally.

For bare-metal, I'm thinking neither `container_linux` nor `flatcar_linux` should actually appear in variable names. Instead of boolean type conditionals, I'd like to see if we can make this more like specifying a channel or flavor once, for better usability. I'd like to borrow the GCP-like semantics of `image = "coreos-stable"`, so I'd aim for something more like `channel = "coreos-stable"` and `channel = "flatcar-stable"`. And then have separate install profiles, its odd to look through a Matchbox store and see Flatcar inside a CL profile. 

Will Flatcar have the same notion of alpha, beta, stable? Interesting the user was kept as "core", but kernel args were renamed.

dongsupark:
@dghubble Thanks.
In the beginning I have thought about creating a different profile `fl` for Flatcar, but I didn't wanted to do that because of the size of PR. But if you think that's the way to go, I can create a different profile.

Setting a flavor (flatcar / coreos) in the channel name sounds like a good idea. I'll update my PR just like you do in https://github.com/poseidon/typhoon/pull/211.

Flatcar has exactly the same channel names, `stable`, `beta`, and `alpha`. Though the default user name is still `core` for now.

dongsupark:
@dghubble Updated, and force-pushed. Is it ok for you?

bare-metal/container-linux/kubernetes/fl/controller.yaml.tmpl
```diff
@@ -0,0 +1,171 @@
+---
```
dghubble: I'd like to keep the Ignition for Container Linux derivatives unified. If I allowed Flatcar to drift, I might have to allow future derivatives and I'd like to keep everything aligned for a while

bare-metal/container-linux/kubernetes/fl/controller.yaml.tmpl
```diff
@@ -0,0 +1,171 @@
...
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes/manifests
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes/cni/net.d
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes/checkpoint-secrets
+        ExecStartPre=/bin/mkdir -p /etc/kubernetes/inactive-manifests
+        ExecStartPre=/bin/mkdir -p /var/lib/cni
+        ExecStartPre=/bin/mkdir -p /var/lib/calico
+        ExecStartPre=/bin/mkdir -p /var/lib/kubelet/volumeplugins
+        ExecStartPre=/usr/bin/bash -c "grep 'certificate-authority-data' /etc/kubernetes/kubeconfig | awk '{print $2}' | base64 -d > /etc/kubernetes/ca.crt"
+        ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/cache/kubelet-pod.uuid
+        ExecStart=/usr/lib/flatcar/kubelet-wrapper \
```
dghubble: You have `/usr/lib/coreos/kubelet-wrapper` in your image, let's keep using that for now.

docs/cl/bare-metal.md
```diff
@@ -186,8 +186,8 @@ module "bare-metal-mercury" {
   # bare-metal
   cluster_name            = "mercury"
   matchbox_http_endpoint  = "http://matchbox.example.com"
-  container_linux_channel = "stable"
-  container_linux_version = "1632.3.0"
+  os_channel              = "stable"
```
dghubble: coreos-stable

bare-metal/container-linux/kubernetes/profiles.tf
```diff
@@ -1,18 +1,39 @@
+locals {
+  flavor  = "${element(split("-", var.os_channel), 0)}"
```
dghubble: One of the nice properties of #211 is that if a user inputs a value like `stable` (instead of `coreos-stable`), the Terraform error strongly hints at the problem showing the AMI data source and indicating that no matching images were found.

I wouldn't do any contortion to make something similar here, but worth considering.

bare-metal/container-linux/kubernetes/profiles.tf
```diff
@@ -1,18 +1,39 @@
...
+  channel = "${element(split("-", var.os_channel), 1)}"
+
+  profile      = "${local.flavor == "flatcar" ? "fl" : "cl"}"
```
dghubble: Sorry. I was unclear in my review. By separate profile, I was referring to separate Matchbox Profiles. A Matchbox profile defines the initrd, kernel, args, etc. that should be used by matched machines. I was proposing instead of using variables throughout the "container-linux-install" "matchbox_profile", I'd consider defining a separate "flatcar-linux-install" profile.

Wasn't proposing making separate directories for cl and fl

bare-metal/container-linux/kubernetes/profiles.tf
```diff
@@ -24,11 +45,11 @@ resource "matchbox_profile" "container-linux-install" {
...
+  template = "${local.template_yaml}"
 
   vars {
-    container_linux_channel = "${var.container_linux_channel}"
-    container_linux_version = "${var.container_linux_version}"
+    os_channel              = "${local.channel}"
+    os_version              = "${var.os_version}"
     ignition_endpoint       = "${format("%s/ignition", var.matchbox_http_endpoint)}"
     install_disk            = "${var.install_disk}"
     container_linux_oem     = "${var.container_linux_oem}"
```
dghubble: Vaguely unrelated to your PR, but this will probably be dropped at some point. So don't worry about the oem setting.

bare-metal/container-linux/kubernetes/profiles.tf
```diff
@@ -40,21 +61,21 @@ data "template_file" "container-linux-install-configs" {
 }
 
 // Container Linux Install profile (from matchbox /assets cache)
-// Note: Admin must have downloaded container_linux_version into matchbox assets.
+// Note: Admin must have downloaded os_version into matchbox assets.
 resource "matchbox_profile" "cached-container-linux-install" {
```
dghubble: You can skip making a flatcar-cached-install profile for now. Without the standard Matchbox setup having an analog of https://github.com/coreos/matchbox/blob/master/scripts/get-coreos, I don't think users will prepare those assets correctly and it'll lead to issues being filed. Let's consider it in a followup.

bare-metal/container-linux/kubernetes/fl/flatcar-linux-install.yaml.tmpl
```diff
@@ -0,0 +1,47 @@
...
+storage:
+  files:
+    - path: /opt/installer
+      filesystem: root
+      mode: 0500
+      contents:
+        inline: |
+          #!/bin/bash -ex
+          curl --retry 10 "${ignition_endpoint}?{{.request.raw_query}}&os=installed" -o ignition.json
+          flatcar-install \
```
dghubble: Hmm, I guess template in coreos vs flatcar in the common container-linux-install.yaml.tmpl (still need to think about renaming the file with all the RH news and shifting names). If the intention is to slowly convert all instances of core to flatcar within the OS, that feels like its going to be painful.

aknuds1:
I'm also interested in Flatcar since I learnt of their fork of Container Linux in the wake of the Red Hat takeover :)

dghubble:
I'm happy to take this on myself. I'd like to get this running myself and I recognize I'm pretty particular about how its implemented, having written Matchbox too. This PR does provide me excellent context for the handful of naming changes and URLs for Flatcar Linux.

dongsupark:
@dghubble Sorry for late reply. I have been on holiday.
I was going to address your review in this week. But if you want to take over this PR, I'd appreciate. Please go ahead. :+1:

dongsupark:
A new PR https://github.com/poseidon/typhoon/pull/220 will replace this PR. So I'll close this one.

PR #211, opened 2018-05-09
Allow Flatcar Linux os_channel on AWS

dghubble:
* Change AWS optional os_channel variable to represent an AMI channel for a Container Linux derivative
* Allow the Flatcar Linux Container Linux derivative os_channel on AWS with `flatcar-stable`, `flatcar-beta`, `flatcar-alpha`. Nice drop-in replacement.
* Require values `stable`, `beta`, `alpha` be changed to `coreos-stable` (default), `coreos-beta`, `coreos-alpha` if they were set

```
$ kubectl describe nodes
...
System Info:
 Machine ID:                 8c859f2bde9e41155bca40d2d85aebb5
 System UUID:                EC2F0AE5-55A4-0FDD-8F06-4F2FC5EA09A3
 Boot ID:                    5d08c248-c559-4687-9bd5-21f79fd03b56
 Kernel Version:             4.14.32-flatcar
 OS Image:                   Flatcar Linux by Kinvolk 1688.5.3 (Rhyolite)
 Operating System:           linux
 Architecture:               amd64
 Container Runtime Version:  docker://17.12.1-ce
 Kubelet Version:            v1.10.2
 Kube-Proxy Version:         v1.10.2
```

### DIFF
--- a/aws/container-linux/kubernetes/ami.tf
+++ b/aws/container-linux/kubernetes/ami.tf
@@ -1,3 +1,12 @@
+locals {
+  # Pick a CoreOS Container Linux derivative
+  # coreos-stable -> Container Linux AMI
+  # flatcar-stable -> Flatcar Linux AMI
+  ami_id = "${local.flavor == "flatcar" ? data.aws_ami.flatcar.image_id : data.aws_ami.coreos.image_id}"
+  flavor = "${element(split("-", var.os_image), 0)}"
+  channel = "${element(split("-", var.os_image), 1)}"
+}
+
 data "aws_ami" "coreos" {
   most_recent = true
   owners      = ["595879546273"]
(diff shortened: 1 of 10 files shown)


dghubble:
Possibly rename the variable to `os_image` to have a really nice alignment with Google Cloud and Digital Ocean which already support `os_image` values `coreos-stable`, `coreos-beta`, etc.` (though no Flatcar yet). Just a uniform way to pick the flavor and channel, no fuss.

dghubble:
rel: #209

aws/container-linux/kubernetes/ami.tf
```diff
@@ -14,6 +23,27 @@ data "aws_ami" "coreos" {
...
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  filter {
+    name   = "name"
+    values = ["Flatcar-${local.channel}-*"]
   }
 }
+
```
dongsupark: I'm not familiar with typhoon & HCL, but am just curious.
Is there a good way to reduce number data definitions?
It seems that a large part of `ami.tf` is defined again in `workers/ami.tf`.

dghubble: Its definitely possible with an output and then consumption of that output.

The main question is whether we _should_. The worker inner module can be used independently for worker pools, so it can't refer to the outer module, so it must have an ami.tf contents. Having the outer kubernetes module use the worker module both to create workers (purpose of the module) _and_ to pick an AMI for controllers feels odd and kinda wrong. Right now, both modules pick an AMI for the instances they create and they happen to pick AMIs in the same way. Favoring some duplication over weird coupling.

dghubble:
Targeting this for the weekend.

Issue #215, opened 2018-05-14
Allow Kubelet API bearer token authentication


dghubble:
### Feature

Today, Kubelet bearer token authentication is not permitted. Allow Kubelet bearer token authentication and enable Webhook authorization (required, as the default would be AlwaysAllow) to enable giving out more fine grained RBAC permissions related to nodes. One example use-case is reducing the Prometheus addon's RBAC privileges from `node/proxy` to `node/metrics`.

This impacts several areas where we need to be careful to avoid regressions:

* kubectl commands
* Prometheus addon
* Potentially metric-server addon

More context: #191

dghubble:
Evaluating PR #191 showed simply adding `--authentication-token-webhook` and `--authorization-mode=Webhook` breaks `kubectl` logs, exec, port-forward, and other essentials (unusual). Some digging shows this seems to be a symptom of not using a special TLS organization in apiserver certificate generation.

Here's how it works.

### Today

Today, Typhoon (along with [bootkube](https://github.com/kubernetes-incubator/bootkube/blob/master/hack/multi-node/user-data.sample#L31) and likely Tectonic)) do not enable Kubelet bearer token authentication. `kubectl` commands like `exec`, `logs`, `port-forward` authenticate with the apiserver and then apiserver uses its X509 `--kubelet-client-certificate` and `--kubelet-client-key` key to authenticate to a particular Kubelet. Requests succeed because the apiserver's certificate authenticates and the Kubelet defaults to `AlwaysAllow` authorization.

Verify the apiserver can present its TLS cert and do whatever it likes:

```
kubectl exec -it kube-apiserver-sr8z2 -n kube-system /bin/bash
curl --cert /etc/kubernetes/secrets/apiserver.crt --key /etc/kubernetes/secrets/apiserver.key  http://NODE-INTERNAL-IP:10250/metrics -k
```

**This design is reasonable and has served well for a while**. No reason to worry. The apiserver X509 certificate is generated upfront and kept safe, since it allows all-access to kubelets.

### Kubelet Bearer Token Auth

Experiment with Kubelet bearer token authentication. Launch a cluster and enable Kubelet bearer token auth on one of the worker nodes by editing the unit file and restarting the kubelet.

```
--authentication-token-webhook
--authorization-mode=Webhook
```

[...]

PR #216, opened 2018-05-14
Allow bearer token authn/authz to the Kubelet

dghubble:
* Require Webhook authorization to the Kubelet
* Switch apiserver X509 client cert org to systems:masters to grant the apiserver admin and satisfy the authorization requirement. kubectl commands like logs or exec that have the apiserver make requests of a kubelet continue to work as before
* https://kubernetes.io/docs/admin/kubelet-authentication-authorization/
* https://github.com/poseidon/typhoon/issues/215


dghubble:
Use-case in the Prometheus addon demonstrated in #217

PR #217, opened 2018-05-14
 Configure Prometheus to scrape Kubelets directly

dghubble:
* Use Kubelet bearer token authn/authz to scrape metrics
* Drop RBAC permission from nodes/proxy to nodes/metrics
* Stop proxying kubelet scrapes through the apiserver, since this required higher privilege (nodes/proxy) and can add load to the apiserver on large clusters


dghubble:
Oh, this is missing `container_*` metrics. We still need a modified metrics path replacement to scrape the Kubelet's CAdvisor metrics path, instead of its default metrics. Will followup with a fix when I get a chance.

dghubble:
Fixed the cadvisor metrics path.

PR #221, opened 2018-05-18
Fix the official docs links

warmchang:

* Fix the official docs links for aws/bare-metal/bare-metal/google-cloud

## Testing

Docs-only, the markdown syntax and hyperlinks are correct.


dghubble:
Thank you! Merged 2ae126bf68362e6fb5e1791909387388a22f766a

PR #224, opened 2018-05-21
Render etcd_initial_cluster using a template_file

bendrucker:
This renders the value for `etcd_initial_cluster` using a different strategy to help work around a terraform bug. The resulting value is unchanged. I found that the `null_resource` setup that's currently present causes errors with the following (simplified) setup:

```hcl
resource "aws_route53_zone" "cluster" {
  name = "kubernetes.cluster"
}

module "typhoon-aws" {
  source = "git::https://github.com/poseidon/typhoon//aws/container-linux/kubernetes"
  dns_zone = "${aws_route53_zone.cluster.name}"
  dns_zone_id = "${aws_route53_zone.cluster.zone_id}"
}
```

The important distinction here is that the DNS zone name and ID are not available in the plan phase and depend on the creation of another resource. In my project, I'm passing in a zone ID and using the `aws_route53_zone` data source in another module to resolve the name for typhoon.

Using this setup, `terraform plan` fails with the following error:

```
Error: Error running plan: 1 error(s) occurred:

* module.typhoon.data.template_file.controller_config: 2 error(s) occurred:

* module.typhoon.data.template_file.controller_config[1]: Resource 'null_resource.repeat' does not have attribute 'triggers.name' for variable 'null_resource.repeat.*.triggers.name'
* module.typhoon.data.template_file.controller_config[0]: Resource 'null_resource.repeat' does not have attribute 'triggers.domain' for variable 'null_resource.repeat.*.triggers.domain'
```

It appears that triggers are only exported as attributes for a `null_resource` when those values are available synchronously. This seems to be a Terraform bug—I'm adding comments to https://github.com/hashicorp/terraform/issues/17641.

Using a `template_file` data source resolves the error on plan and results in a working project. While it's a little verbose to have to re-specify `vars {}`, I also think it's easier to understand.

### DIFF
--- a/aws/container-linux/kubernetes/controllers.tf
+++ b/aws/container-linux/kubernetes/controllers.tf
@@ -54,7 +54,7 @@ data "template_file" "controller_config" {
     etcd_domain = "${var.cluster_name}-etcd${count.index}.${var.dns_zone}"
 
     # etcd0=https://cluster-etcd0.example.com,etcd1=https://cluster-etcd1.example.com,...
-    etcd_initial_cluster = "${join(",", formatlist("%s=https://%s:2380", null_resource.repeat.*.triggers.name, null_resource.repeat.*.triggers.domain))}"
+    etcd_initial_cluster = "${join(",", data.template_file.etcds.*.rendered)}"
 
     kubeconfig            = "${indent(10, module.bootkube.kubeconfig)}"
     ssh_authorized_key    = "${var.ssh_authorized_key}"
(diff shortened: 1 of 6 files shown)


dghubble:
Initially looks nice. And we can do away with the null_resource trigger hack. I'll target this for v1.10.4 time frame and revisit in a few days - v1.10.3 is already underway.

bendrucker:
Sounds good, thanks!

dghubble:
Rebased and merged as 6a581ab5779a999c8575c30b46f0868b0a52cfaf

Thanks!

dghubble:
This change broken Google Cloud container-linux and fedora-atomic so it'll be partially reverted in https://github.com/poseidon/typhoon/pull/237/commits/6676484490ae70753045aeeaef4939c4d18a9803

Issue #140, opened 2018-02-19
Introduce Container Linux config additive customizations


dghubble:
### Feature

Introduce the ability for users to supply Container Linux Config "snippets" for controllers or workers. Snippets are valid Container Linux Configs which are additively merged into the base Container Linux Configs that Typhoon uses to provision controllers and workers. Configs are validated, merged into a single config for use in instance user-data. Validation errors are shown with line numbers during `terraform plan`.

This is a major feature and has been long awaited.

* Allows users to add custom systemd units, network configs, files, raid arrays, or other customizations related to on-host disk provisioning as supported by Container Linux Configs (at Ignition v2.0.0).
* Reduces user temptation to add a Terraform variable for every option. This helps keep Typhoon modules readable and maintainable over the long run.
* See the Container Linux Config [spec](https://coreos.com/os/docs/latest/configuration.html) and [examples](https://coreos.com/os/docs/latest/clc-examples.html)

I've made changes to the `terraform-provider-ct` provider plugin in v0.2.1 to enable this feature.

### Caveats:

* There are limits to what additive merging (technically `append`) will allow. You cannot remove bits of the base config added by Typhoon. This is part of the reason Typhoon focuses on minimalism.
* Typhoon cannot make guarantees about different end-user customizations. We'll continue to design, build, and test clusters as usual, but cannot entertain reports concerning end-user's customizations (e.g. "I customized X and now Y is broken").

### Components

- [x] Introductory Docs
- [x] Update "customization" docs. We now have a better story.
- [x] Guide for updating terraform-provider-ct plugin to v0.2.1
- [x] AWS
- [x] Digital Ocean
- [x] Google Cloud
- [ ] Bare-Metal

### Problems

* ~`terraform-provider-ct` is not currently an official terraform provider. This actually presents substantial migration difficulties. Admins who already manage a bunch of clusters and then swap the provider binary on their system will notice diff's to those existing clusters. Applying those diffs can destroy clusters as terraform believes its task is to replace every node. I'm still investigating this.~

dghubble:
The issue with terraform-provider-ct [patch#22](https://github.com/coreos/terraform-provider-ct/pull/22) causing Terraform diffs was determined to be related to the provider updating the Ignition version. See the [full description](https://github.com/coreos/terraform-provider-ct/pull/22#issuecomment-368062792). 

[...]

PR #245, opened 2018-06-15
Container Linux Config snippets for Bare-Metal platform

hcbraun:
same like in the AWS platform
requires the ct provider: https://github.com/coreos/terraform-provider-ct


dghubble:
In #140 I laid out how I'd like to support heterogeneous nodes instead of this approach.

dghubble:
Closed by #269

PR #254, opened 2018-06-27
Update google-cloud.md

ballPointPenguin:
including a providers block in .terraformrc has been deprecated in favor of adding the plugin directly to .terraform.d/plugins/

upgarde to terraform-provider-ct-v0.3.0

* Specific change
* Specific change

## Testing


dghubble:
Terraform still allows the binary to be placed wherever you like on the system. If this change is to be made now, it needs to be done everywhere and explained to users.

It'll probably be clearer and less disruptive to just wait until the next Terraform version. Folks will be mentally prepared for minor migration annoyances like this to with the perk being they'll get [insert features of next Terraform version].

PR #269, opened 2018-07-26
Support Container Linux Config snippets on bare-metal

### LINKED ISSUE #140: Introduce Container Linux config additive customizations

dghubble:
Introduce [Container Linux Config snippets](https://typhoon.psdn.io/advanced/customization/#container-linux) on bare-metal. These were introduced for AWS/Google/DO previously.
  * Validate and additively merge custom Container Linux Configs during terraform plan
  * Define files, systemd units, dropins, networkd configs, mounts, users, and more
  * [Require](https://typhoon.psdn.io/cl/bare-metal/#terraform-setup) `terraform-provider-ct` plugin v0.2.1 (action required!)

Container Linux config snippets provide a mechanism for adding host customizations and enforcing those customizations merge to produce valid Ignition. It also eliminates advanced use cases that previously required forking (just to name a few from my own clusters - custom RAID setups, bonded NICs, custom mount units). The actual snippets are passed as raw string contents, meaning they can be read from files or prepared by the user in some way they see fit (its not a Typhoon concern).

## Usage

Bare-Metal clusters allow different Container Linux snippets to be used for each node (since hardware may be heterogeneous). Define the optional `clc_snippets` map variable using controller or worker keys.

```
# EDITED
module "bare-metal-mercury" {
  ...
  controller_names = ["node1"]
  worker_names = [
    "node2",
    "node3",
  ]
  clc_snippets = {
    "node2" = [
      "${file("./hello.yaml")}"
    ]
    "node3" = [
      "${file("./world.yaml")}",
      "${file("./hello.yaml")}",
    ]
  }
  ...
}
```

```
storage:
  files:
    - path: /opt/hello
      filesystem: root
      contents:
        inline: |
          Hello World
      mode: 0644
```

Edge case: If you wish to define snippets for just a few machines, only add mappings for those machines. There is **no need** to do `"node3" = []` for example, just leave it out. Empty list is not the right default due to some complicated defaulting choices forced by HCL constraints.

### Tradeoffs

* Unfortunately, this means every bare-metal user will need to install the terraform-provider-ct plugin. Even if they don't intend to use the Container Linux config snippet advanced feature. I'm not thrilled doing this, but at the same time, that plugin is required for all the cloud variants of Typhoon. So maybe its not too unreasonable.
* Users who do need to apply the exact same snippet to all machines will need to use a map with a key for each machine to a list containing that snippet. Its repetitive, but I think this is acceptable to abilities to control snippets per-machine.

### Followups

Bare-Metal variables `controller_networkds` and `worker_networkds` that were marked as "unofficial, undocumented, unsupported, temporary" can be deleted.

Closes #140 #245

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,15 @@ Notable changes between versions.
 
 ## Latest
 
+#### Bare-Metal
+
+* Introduce [Container Linux Config snippets](https://typhoon.psdn.io/advanced/customization/#container-linux) on bare-metal
+  * Validate and additively merge custom Container Linux Configs during terraform plan
+  * Define files, systemd units, dropins, networkd configs, mounts, users, and more
+  * [Require](https://typhoon.psdn.io/cl/bare-metal/#terraform-setup) `terraform-provider-ct` plugin v0.2.1 (action required!)
+
+## v1.11.1
+
 * Kubernetes [v1.11.1](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#v1111)
 
 #### Addons
--- a/bare-metal/container-linux/kubernetes/profiles.tf
+++ b/bare-metal/container-linux/kubernetes/profiles.tf
@@ -118,8 +118,17 @@ resource "matchbox_profile" "flatcar-install" {
 resource "matchbox_profile" "controllers" {
   count                  = "${length(var.controller_names)}"
   name                   = "${format("%s-controller-%s", var.cluster_name, element(var.controller_names, count.index))}"
-  container_linux_config = "${element(data.template_file.controller-configs.*.rendered, count.index)}"
+  raw_ignition = "${element(data.ct_config.controller-ignitions.*.rendered, count.index)}"
 }
+
+data "ct_config" "controller-ignitions" {
+  count = "${length(var.controller_names)}"
+  content = "${element(data.template_file.controller-configs.*.rendered, count.index)}"
+  pretty_print = false
+  # Must use direct lookup. Cannot use lookup(map, key) since it only works for flat maps
+  snippets = ["${local.controller_clc_map[element(var.controller_names, count.index)]}"]
+}
+
 
 data "template_file" "controller-configs" {
   count = "${length(var.controller_names)}"
@@ -143,7 +152,16 @@ data "template_file" "controller-configs" {
 resource "matchbox_profile" "workers" {
   count                  = "${length(var.worker_names)}"
   name                   = "${format("%s-worker-%s", var.cluster_name, element(var.worker_names, count.index))}"
-  container_linux_config = "${element(data.template_file.worker-configs.*.rendered, count.index)}"
+  raw_ignition = "${element(data.ct_config.worker-ignitions.*.rendered, count.index)}"
+}
+
+
+data "ct_config" "worker-ignitions" {
+  count = "${length(var.worker_names)}"
+  content = "${element(data.template_file.worker-configs.*.rendered, count.index)}"
+  pretty_print = false
+  # Must use direct lookup. Cannot use lookup(map, key) since it only works for flat maps
+  snippets = ["${local.worker_clc_map[element(var.worker_names, count.index)]}"]
 }
 
 data "template_file" "worker-configs" {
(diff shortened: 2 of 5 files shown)


dghubble:
This can be made simpler still - the maps `controller_clc_snippets` and `worker_clc_snippets` can be consolidated into one input variable `clc_snippets`. Using a different name has the advantage that it can't be confused with the controller/worker _lists_ used for cloud providers - bare-metal supports snippets per-node because its legitimately quite different from clouds

bugre:
I'm trying to implement a mini typhon cluster with 3 servers, on bare-metal. I'm using:
14:31 $ terraform -v providers
Terraform v0.12.21
+ provider.ct v0.4.0
+ provider.local v1.4.0
+ provider.matchbox v0.3.0
+ provider.null v2.1.2
+ provider.template v2.1.2
+ provider.tls v2.1.1

In order to personalize the network section of each server, to assign static IP's and distinct bonding of multiple NICs. I'm trying to do it using the clc_snippets. I have the snippets templates, they get correctly included to each server's generated ignition file, but the variables don't get replaced as I had expected.

As i understood it from the PR #269 and issue #140 , i should use terraform variables. So i tried adding ${var.net-if-ip} inside the snippets where the custom values must be, but to no result. The content is not replaced with the value of the variable 'net-if-ip' declared in variables.tf and myconfig.tfvars.  

Please, could someone, point me into a direction with some more docs or example, where the clc_snippets contain replacement variables? Or did i understand it completely wrong? Should i be using some different approach to archive the personalized 'network' and other units in the config of each server?

I've archived a similar solution using only terraform with matchbox templating, but i would like to "upgrade" the solution to use thyphoon's model ,that would help me resolve other issues i have, plus allowing the install of kubernetes without aditional effort.

If i should open an issue (i did not as this is probably me not knowing how to do it)

Thanks

dghubble:
`clc_snippets` is a map from machine names to valid Container Linux Configs. It's not related to templating and certainly won't use Terraform inputs inside them. See the [docs](https://typhoon.psdn.io/advanced/customization/#container-linux).

However, the Container Linux Configs are passed as content directly. They don't necessarily need to be directly from files, you can prepare those configs, however you like, using any Terraform mechanisms you like, outside of Typhoon.

bugre:
@dghubble , your second paragraph did helpe me understand that i could use a different terraform data sources/templatefile structures that dynamically populate a "variable" with the CLC snipped content. 
Thanks for taking the time to answer.

PR #275, opened 2018-08-10
Upgrade terraform-provider-ct to v0.3.0 to provide ignition schema v2.2.

cskarby:
## Warning

It is **not safe** to upgrade the terraform-provider-ct plugin between
minor versions on existing managed clusters. Minor version upgrades of this
plugin **will re-provision your cluster** because the Ignition config output has
changed.

However Typhoon explicitly [encourages using a replacement strategy rather than inplace upgrades](https://typhoon.psdn.io/topics/maintenance/#upgrades), which could make it acceptable to upgrade the terraform-provider-ct plugin with a proper warning in documentation.

## Changes

Upgraded documentation. The user must upgrade the plugin manually by overwriting the old plugin.

## Testing

terraform-provider-ct [v0.3.0](https://github.com/coreos/terraform-provider-ct/releases/tag/v0.3.0) is upgraded to using container-linux-config-transpiler [v0.8.0](https://github.com/coreos/container-linux-config-transpiler/releases/tag/v0.8.0). Ran the config transpiler on the raw cl templates, and got the following warnings. The errors in bare-metal is due to the transpiler not understanding the terraform template language (the plugin passes the result from a rendered template to the transpiler.) 

```
./aws/container-linux/kubernetes/cl/controller.yaml.tmpl
warning at line 126, column 7
mode unspecified for file, defaulting to 0644

./aws/container-linux/kubernetes/workers/cl/worker.yaml.tmpl
warning at line 96, column 7
mode unspecified for file, defaulting to 0644

./bare-metal/container-linux/kubernetes/cl/controller.yaml.tmpl
error: yaml: unmarshal errors:
  line 167: cannot unmarshal !!str `${netwo...` into types.Networkd
Failed to parse config

./bare-metal/container-linux/kubernetes/cl/install.yaml.tmpl
./bare-metal/container-linux/kubernetes/cl/worker.yaml.tmpl
error: yaml: unmarshal errors:
  line 100: cannot unmarshal !!str `${netwo...` into types.Networkd
Failed to parse config

./digital-ocean/container-linux/kubernetes/cl/controller.yaml.tmpl
warning at line 132, column 7
mode unspecified for file, defaulting to 0644

./digital-ocean/container-linux/kubernetes/cl/worker.yaml.tmpl
warning at line 102, column 7
mode unspecified for file, defaulting to 0644

./google-cloud/container-linux/kubernetes/cl/controller.yaml.tmpl
warning at line 127, column 7
mode unspecified for file, defaulting to 0644

./google-cloud/container-linux/kubernetes/workers/cl/worker.yaml.tmpl
warning at line 97, column 7
mode unspecified for file, defaulting to 0644
```

Removing the terraform variable also renders the rest of the bare-metal cl templates without errors.
```
./bare-metal/container-linux/kubernetes/cl/controller.yaml.tmpl
warning at line 133, column 7
mode unspecified for file, defaulting to 0644

./bare-metal/container-linux/kubernetes/cl/worker.yaml.tmpl
warning at line 94, column 7
mode unspecified for file, defaulting to 0644
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 Notable changes between versions.
 
+* [Recommend](https://typhoon.psdn.io/cl/bare-metal/#terraform-setup) updating `terraform-provider-ct` plugin from v0.2.1 to [v0.3.0](https://github.com/coreos/terraform-provider-ct/releases/tag/v0.3.0) (action recommended)
+
 ## v1.11.2
 
 * Kubernetes [v1.11.2](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#v1112)
--- a/docs/cl/aws.md
+++ b/docs/cl/aws.md
@@ -24,9 +24,9 @@ Terraform v0.11.1
 Add the [terraform-provider-ct](https://github.com/coreos/terraform-provider-ct) plugin binary for your system.
 
 ```sh
-wget https://github.com/coreos/terraform-provider-ct/releases/download/v0.2.1/terraform-provider-ct-v0.2.1-linux-amd64.tar.gz
-tar xzf terraform-provider-ct-v0.2.1-linux-amd64.tar.gz
-sudo mv terraform-provider-ct-v0.2.1-linux-amd64/terraform-provider-ct /usr/local/bin/
+wget https://github.com/coreos/terraform-provider-ct/releases/download/v0.3.0/terraform-provider-ct-v0.3.0-linux-amd64.tar.gz
+tar xzf terraform-provider-ct-v0.3.0-linux-amd64.tar.gz
+sudo mv terraform-provider-ct-v0.3.0-linux-amd64/terraform-provider-ct /usr/local/bin/
 ```
 
 Add the plugin to your `~/.terraformrc`.
--- a/docs/cl/bare-metal.md
+++ b/docs/cl/bare-metal.md
@@ -124,9 +124,9 @@ sudo mv terraform-provider-matchbox-v0.2.2-linux-amd64/terraform-provider-matchb
 Add the [terraform-provider-ct](https://github.com/coreos/terraform-provider-ct) plugin binary for your system.
 
 ```sh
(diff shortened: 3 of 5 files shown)


dghubble:
This isn't safe or complete.

I wrote a [full warning](https://github.com/coreos/terraform-provider-ct/pull/22#issuecomment-368062792) for plugin versioning with `terraform-provider-ct` a while ago. Clusters created when using the v0.2.x plugins should continue using the v0.2.x plugins, those created using the v0.3.x plugins should continue using the v0.3.x version. `terraform-provider-ct` is an unofficial plugin. As such, Terraform treats it globally - each cluster isn't allowed to have its own version of the dependency (as you'd expect with say the `aws` provider). Use of blue/green deployment does not save you from this problem.

Here, you're updating a single cluster. That's fine. Its the simple case. But users who have existing clusters created using the `v0.2.x` plugin, if they update to `v0.3.0`, that means running `terraform apply` again for **any** of those clusters managed on the same system is a change. It violates an important invariant - if cluster's definition hasn't changed (i.e. no change to `source=...`, no change to fields) a `terraform apply` is a NoOp. That also means if they did have a legitimate change (e.g. add new workers) they can't apply it until all bare-metal clusters use the new plugin (recreation) - its a non-starter and serious risk for impacting all managed clusters at one.

Ultimately, this will require a much better migration story (here was the [plan](https://github.com/poseidon/typhoon/blob/d8d524d10b1a3b5f0310fd2718f11caafecff51d/docs/topics/maintenance.md#terraform-provider-ct-v021) just for v0.2.0 to v0.2.1). Likely, in upcoming releases, bare-metal clusters will ignore post-creation alterations to `user-data`. Then enough time must pass that most existing clusters are on a version with that change. Then we'll be able to write docs on updating the global plugin without affecting all managed clusters. Alternately, Terraform allows custom plugins to be treated non-globally like official plugins, but that's pretty unlikely.

Regarding the templating issue - templates should be fully evaluated before being used by `ct_config`. I think this is an artifact of how you're testing, since it should be the case when using the module.

However: If you're a **new user**, you may feel free to start out using `terraform-provider-ct` `v0.3.0` though. It doesn't contain much that's new, but its not harmful.

PR #287, opened 2018-08-22
Fix firewall to allow etcd client traffic between controllers

dghubble:
* Broaden internal-etcd firewall rule to allow etcd client traffic (2379) from other controller nodes
* Previously, kube-apiservers were only able to connect to their node's local etcd peer. While master node outages were tolerated, reaching a healthy peer took longer than necessary in some cases
* Reduce time needed to bootstrap a cluster

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@ Notable changes between versions.
 
 #### Google Cloud
 
+* Fix firewall to allow etcd client port 2379 traffic between controller nodes ([#287](https://github.com/poseidon/typhoon/pull/287))
+  * kube-apiservers were only able to connect to their node's local etcd peer. While master node outages were tolerated, reaching a healthy peer took longer than neccessary in some cases
+  * Reduce time needed to bootstrap the cluster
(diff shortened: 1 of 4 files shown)


dghubble:
This was most evident from `kubectl get cs` (ignore scheduler and controller manager) several times on a GCP cluster. Notice, whichever apiserver handles the request, you'll see it views its etcd peer as healthy.

```
etcd-0               Healthy     {"health":"true"}                                                                                                                                            
etcd-2               Unhealthy   Get https://yavin-etcd2.domain.com.:2379/health: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)   
etcd-1               Unhealthy   Get https://yavin-etcd1.domain.com.:2379/health: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

With the fix, the results align with other multi-master clusters.

```
etcd-0               Healthy     {"health":"true"}                                                                                                                                            
etcd-1               Healthy     {"health":"true"}                                                                                                                                            
etcd-2               Healthy     {"health":"true"}                                                                                                                                            
```

Note, you can still shutdown masters if a quorum of nodes are up and expect `kubectl` to work. So this doesn't directly impact availability, but was definitely undesired.

Issue #294, opened 2018-09-15
ingress not working with Digitalocean


naarani:
## Bug

Ingress configuration to expose a service not working with addon/nginx/digitalocean

### Environment

* Platform: digital-ocean
* OS: container-linux
* Terraform: last

### Problem

If I use the addon the default backed works, however I can't add an ingress definition to expose a service. The only way is to removed the addon and install default kubernetes nginx ingress and exposing IP ports on each node

### Steps to Reproduce
file ingress-test.yaml:

apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-test
spec:
  tls:
    - hosts:
      - myhost-test.com
      secretName: my-tls
  rules:
  - host: myhost-test.com
    http:
      paths:
      - backend:
          serviceName: demo
          servicePort: 8080

dghubble:
What specifically is the problem you're seeing (i.e. error message, bug, logs, etc.). Still running fine for me.

naarani:
instead of a response from the service, I got a response only from default backend:
I've tried to track it but I've find no errors

Only changing the Nginx the file work:
did you have an example of a running ingress for a service?

thx

dghubble:
This is almost certainly something wrong with your application, your TLS, or your Ingress resource. Walk through https://kubernetes.io/docs/concepts/services-networking/ingress/ If its your first usage, consider just an HTTP app first, then introduce TLS termination. Ensure your `my-tls` secret is formatted in exactly the way its supposed to be and in the right namespace. Tail the logs of the Ingress controller to verify it notices your Ingress resource. Make sure `myhost-test.com` (or whatever) resolves to a worker, possibly using a CNAME to the convenience record created with the cluster for this purpose https://typhoon.psdn.io/addons/ingress/#digital-ocean

dghubble:
Also, HTTP services should be using port 80. Pods can use 8080 if you want, but the service is 80 or 443 (if it does its own termination).

naarani:
thanks for the informations, I should have done something wrong, I'll need to check my configuration

PR #295, opened 2018-09-15
Update daemonset.yaml

naarani:
Added a configmap and attached it to Nginx ingress to controll some default parameters.

HSTS add security, but having it installed by default is worst then the problems it is called to resolve:
it's an advanced parameters and people should set it only if they know how it works to prevent problems.
 hsts-include-subdomains: "false"
 hsts-max-age: "0"
 hsts: "false"

Hide some some server information, like the name and version, it's useful to reduce attack exposition of the services.
proxy-hide-headers: "Server"
server-tokens: "False"

reference to parameters
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/
https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/customization

* Specific change
* Specific change

## Testing

### DIFF
--- a/addons/nginx-ingress/digital-ocean/daemonset.yaml
+++ b/addons/nginx-ingress/digital-ocean/daemonset.yaml
@@ -1,3 +1,20 @@
+---
+
+kind: ConfigMap
+apiVersion: v1
+data:
+  hsts-include-subdomains: "false"
+  hsts-max-age: "0"
+  hsts: "false"
+  proxy-hide-headers: "Server"
+  server-tokens: "False"
+metadata:
+  name: nginx-configuration
+  namespace: ingress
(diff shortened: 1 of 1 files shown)


dghubble:
Most of these values aren't any different from the defaults (except `server-tokens`), you've only applied it for DigitalOcean, and this isn't following the one-manifest-per-file convention like all other addons. Overall, this doesn't seem very compelling and folks can add whatever Nginx customizations are desired on their own, addon minutiae aren't a Typhoon concern.

dghubble:
Wait, you're the same person having difficulties using the ingress controller on DigitalOcean in https://github.com/poseidon/typhoon/issues/294.

PR #306, opened 2018-10-02
Require terraform-provider-digitalocean plugin ~> 1.0

dghubble:
* Require a terraform-provider-digitalocean plugin version of 1.0 or higher within the same major version (e.g. allow 1.1 but not 2.0)
* Without this change, using the new digitalocean [1.0](https://github.com/terraform-providers/terraform-provider-digitalocean/blob/master/CHANGELOG.md#100-september-27-2018) provider wouldn't be allowed since its the next major version

Note: I'm intentionally not using `>= 0.1.2` since that would allow using a 2.0. When the next digitalocean major version is released, we'll re-validate and alter the constraint.

## Testing

I've tested with the digitalocean 1.0 plugin and it creates the resources for a Typhoon Kubernetes cluster as before.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@ Notable changes between versions.
 * Update CoreDNS from 1.1.3 to 1.2.2
 * Update Calico from v3.2.1 to v3.2.3
 
+#### DigitalOcean
+
+* Require terraform-provider-digitalocean plugin 1.0 (or higher but < 2.0)
+
 #### Addons
 
 * Update Prometheus from v2.3.2 to v2.4.2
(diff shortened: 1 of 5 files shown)


dghubble:
DigitalOcean in behaving in a flaky way, creating some VMs successfully, while others just hang in their UI. Perhaps they're running out of addressed in NYC3. Unfortunately, API hiccups and outages aren't uncommon. I'll try validating this later.

dghubble:
Looks like there was determined to be a DigitalOcean [incident](https://status.digitalocean.com/incidents/988j7ydm3kh8).

PR #314, opened 2018-10-15
support IOPS in AWS (required for io1)

rfairburn:
This change allows for specifying a disk_type of io1 in AWS.  Without having IOPS as a configurable variable, `terraform apply` will error out with a missing required setting.

* added disk_iops to both Container Linux and Fedora Atomic AWS options
* set disk_iops to a default of 120 (default iops for the 40GB/gp2 settings)

## Testing

I was able to successfully provision both workers and controllers as io1 with a custom number of iops after this change.

### DIFF
--- a/aws/container-linux/kubernetes/controllers.tf
+++ b/aws/container-linux/kubernetes/controllers.tf
@@ -30,6 +30,7 @@ resource "aws_instance" "controllers" {
   root_block_device {
     volume_type = "${var.disk_type}"
     volume_size = "${var.disk_size}"
+    iops        = "${var.disk_iops}"
   }
 
   # network
--- a/aws/container-linux/kubernetes/variables.tf
+++ b/aws/container-linux/kubernetes/variables.tf
@@ -59,6 +59,12 @@ variable "disk_type" {
   description = "Type of the EBS volume (e.g. standard, gp2, io1)"
 }
 
+variable "disk_iops" {
+  type        = "string"
+  default     = "0"
+  description = "IOPS of the EBS volume (required for io1)"
+}
+
 variable "worker_price" {
   type        = "string"
   default     = ""
--- a/aws/container-linux/kubernetes/workers/variables.tf
+++ b/aws/container-linux/kubernetes/workers/variables.tf
@@ -52,6 +52,12 @@ variable "disk_type" {
   description = "Type of the EBS volume (e.g. standard, gp2, io1)"
 }
 
+variable "disk_iops" {
+  type        = "string"
+  default     = "0"
+  description = "IOPS of the EBS volume (required for io1)"
+}
+
 variable "spot_price" {
   type        = "string"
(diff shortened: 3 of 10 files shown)


aws/container-linux/kubernetes/controllers.tf
```diff
@@ -30,6 +30,7 @@ resource "aws_instance" "controllers" {
   root_block_device {
     volume_type = "${var.disk_type}"
     volume_size = "${var.disk_size}"
+    iops        = "${var.disk_iops}"
```
dghubble: Begs the question of whether `iops` is a valid parameter for standard and gp2 disks and if so, what the value should be.

aws/container-linux/kubernetes/variables.tf
```diff
@@ -59,6 +59,12 @@ variable "disk_type" {
   description = "Type of the EBS volume (e.g. standard, gp2, io1)"
 }
 
+variable "disk_iops" {
+  type        = "string"
+  default     = "120"
+  description = "IOPS of the EBS volume (required for io1)"
```
dghubble: Just reading docs, "0 If the volume is not a provisioned IOPS image, otherwise the supported IOPS count". So if we choose to permit it, this shouldn't default to 120.

aws/container-linux/kubernetes/workers/variables.tf
```diff
@@ -52,6 +52,12 @@ variable "disk_type" {
   description = "Type of the EBS volume (e.g. standard, gp2, io1)"
 }
 
+variable "disk_iops" {
+  type        = "string"
+  default     = "120"
```
dghubble: Just pulled from the console I take it?

rfairburn:
I will look and see if there is a better way to handle iops and default values.  Ideally we would conditionally omit the iops setting, but Terraform does not appear that it will support this directly until 0.12 (https://www.hashicorp.com/blog/terraform-0-12-conditional-operator-improvements).

dghubble:
If I pass `0`, it seems to set the default it normally would for gp2. Haven't tried standard. That's likely a better default if AWS will automatically use the right value based on size and type. Then only the folks with `io` need to consider it and they're opting into that more advanced experience.

rfairburn:
I updated the default to 0 since AWS appears to do the right thing with that value set.  Please let me know if you have any further concerns.

dghubble:
Sounds alright. I'll give this another look soon. Already switched over all clusters to v1.12.1 about 2 days back so it'll miss the v1.12.1 train, but can make it into the next release.

dghubble:
Can you squash to one commit and rebase please?

rfairburn:
Done

dghubble:
Rebased on master and added a changelog entry. Merged as 0be4673e44ecb28758cf292968674603b1c1028e.

Thanks!

PR #301, opened 2018-09-22
Update Kubernetes from v1.11.3 to v1.12.x

dghubble:
Trial the release candidates.

* Control plane changes https://github.com/poseidon/terraform-render-bootkube/pull/77

Known issues:

* https://github.com/kubernetes/kubernetes/issues/68973 (using a workaround)
* ~https://github.com/kubernetes/kubernetes/issues/68986~ (submitted a fix that's in v1.12.1 now)
* Kubelet doesn't start the pod-checkpointer to restore control plane on power cycling https://github.com/kubernetes-incubator/bootkube/issues/1001 (fix in progress, confirmed)

PR #317, opened 2018-10-16
Add docker/default seccomp to control plane and addons

dghubble:
* Annotate pods, deployments, and daemonsets to start containers with the Docker runtime's default seccomp profile
* Overrides Kubernetes default behavior which started containers with seccomp=unconfined
* https://docs.docker.com/engine/security/seccomp/#pass-a-profile-for-a-container

rel: https://github.com/poseidon/terraform-render-bootkube/commit/2437023c1050609b749850e9b2301a6f00713680


dghubble:
Conformance test runs for v1.12 were already rebased atop the terraform-render-bootkube seccomp change so the control plane change has gone through conformance testing. #301

No issues running addons either in my testing. As expected since none of these apps should be making any of the restricted syscalls in the default docker seccomp profile.

Issue #322, opened 2018-10-18
Disable Kubelet unauthenticated read-only port


elemental-lf:
## Bug

Not a bug per se but a  leakage of possibly sensitive information

### Environment

All environments AFAIK.

### Problem

While researching another issue I just came across the fact that ``kubelet`` opens up an unauthenticated HTTP endpoint on port 10255 which can be used to get a dump of all pod manifests (``/pods``). I just checked on my Typhoon based cluster and the port is indeed open and reachable from the outside. Was this a conscious decision or is this port even used by something at the moment? If not, I'd like to suggest calling ``kubelet`` with ``--read-only-port 0`` to disable this unauthenticated access to possibly sensitive information. I've disabled this port for one ``kubelet`` of my cluster and haven't seen any ill side effects, yet.
Some context:
- https://github.com/kubernetes/kubernetes/pull/59666#issuecomment-373829976
- https://github.com/kubernetes/kubeadm/issues/732

### Desired Behavior

If nothing precludes this I think this port should be closed.

### Steps to Reproduce

``curl http://$K8S_HOST:10255/pods``

dghubble:
correction: Typhoon AWS began using an apiserver TCP health check by v1.9.4.

So Typhoon v1.10.2 (April 2018) achieved apiserver load balancing without requiring the read-only port. But the read-only port was still used for Prometheus to scrape Kubelet and cAdvisor metrics and for Heapster to collect metrics (i.e. my attempts in May).

Typhoon v1.10.3 added Kubelet [bearer token authentiation](https://github.com/poseidon/typhoon/pull/216) and switched Prometheus to scrape Kubelet and cAdvisor [via the secure API](https://github.com/poseidon/typhoon/pull/217).

Typhoon v1.12.1 still requires the read-only port for Heapster addon (or metrics-server). You'll notice `kubectl top` commands failing if its removed (or if you remove the firewall rules allowing the traffic). But today, Heapster has [options](https://github.com/kubernetes/heapster/blob/da5088644cb0fdebe54d65bfc2976708d4c850e9/docs/source-configuration.md) to use the kubelets' secure port.  While the Kubernetes default `system:heapster` ClusterRole lacks the correct GET access to node/stats (see https://github.com/kubernetes/heapster/issues/1936), that's simple enough to patch in the Heapster addon.

dghubble:
In the next release, the kubelet read-only port is no longer needed.

elemental-lf:
[...]

PR #323, opened 2018-10-19
Configure Heapster to source metrics from Kubelet authenticated API

dghubble:
* Heapster can now get nodes (i.e. kubelets) from the apiserver and source metrics from the Kubelet authenticated API (10250) instead of the Kubelet HTTP read-only API (10255)
* https://github.com/kubernetes/heapster/blob/master/docs/source-configuration.md
* Use the heapster service account token via Kubelet bearer token authn/authz.
* Permit Heapster to skip CA verification. The CA cert does not contain IP SANs and cannot since nodes get random IPs that aren't known upfront. Heapster obtains the node list from the apiserver, so the risk of spoofing a node is limited. For the same reason, Prometheus scrapes must skip CA verification for scraping Kubelet's provided by the apiserver.
* https://github.com/poseidon/typhoon/blob/v1.12.1/addons/prometheus/config.yaml#L68
* Create a heapster ClusterRole to work around the default Kubernetes
`system:heapster` ClusterRole lacking the proper GET `nodes/stats`
access. See https://github.com/kubernetes/heapster/issues/1936

rel: #322

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,10 @@ Notable changes between versions.
 
 * Add `disk_iops` variable for EBS volume IOPS ([#314](https://github.com/poseidon/typhoon/pull/314))
 
+#### Addons
+
+* Configure Heapster to scrape Kubelets with bearer token auth ([#323](https://github.com/poseidon/typhoon/pull/323))
+
 ## v1.12.1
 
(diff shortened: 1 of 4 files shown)


dghubble:
This change should be the final piece to disable the Kubelet read-only port!

PR #325, opened 2018-10-19
Consistify naming for AWS and Digital Ocean Ignition data sources

tuhtah:
Small nitpicks to follow the same data source naming convention for all platforms.

Do unfortunately not have an AWS nor Digitial Ocean setup available for testing, so please verify accordingly before accepting.


dghubble:
Rebased and merged as bfa1a679ebff19a3ad5cf9b0525dac6a2ab31fda with some followup naming/commenting fixes. Thanks!

PR #337, opened 2018-11-04
Set kube-apiserver kubelet preferred address types

### LINKED ISSUE #186: DO/bare-metal: Change preferred-address-type instead of overriding hostname
## Feature Request

On DO/bare-metal we currently override the hostname with the private ip (relevant [commit](https://github.com/poseidon/typhoon/commit/097dcdf47ef1fed39e184842926955cbdd833d0c)), as the hostname isn't resolvable:
> * Kubelets should register nodes via their private IPv4 address,
as provided by the metadata service from Digital Ocean
> * By default, Kubelet exec's hostname to determine the name it should
use when registering with the apiserver. On Digital Ocean, the hostname
is not routeable by other instances. Digital Ocean does not run an
internal DNS service.
> * Fixes issue where the apiserver can't reach the worker nodes. This
prevented kubectl logs and exec commands from working

### Feature

It would be more optimal to use the real hostname as it is easier to identify a node by its hostname, rather than its ip.
We can archive that by changing (`kube-apiserver`:
>       --kubelet-preferred-address-types strings                 List of the preferred NodeAddressTypes to use for kubelet connections. (default [Hostname,InternalDNS,InternalIP,ExternalDNS,ExternalIP])

Which change how the apiserver connect to the nodes. I'm currently using `--kubelet-preferred-address-types=InternalIP,Hostname,ExternalIP` on a regular `bootkube` DO cluster without any issues.

### Tradeoffs

I'm not sure how this will affect AWS/GCE clusters.

dghubble:
* Prefer InternalIP and ExternalIP over the node's hostname, to match upstream behavior and kubeadm
* Previously, hostname-override was used to set node names to internal IP's to work around some cloud providers not resolving hostnames for instances (e.g. DO droplets)

Closes #186

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@ Notable changes between versions.
 
 ## Latest
 
+* Change kube-apiserver `--kubelet-preferred-address-types` to InternalIP,ExternalIP,Hostname
+
 ## v1.12.2
 
 * Kubernetes [v1.12.2](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#v1122)
--- a/README.md
+++ b/README.md
@@ -87,10 +87,10 @@ In 4-8 minutes (varies by platform), the cluster will be ready. This Google Clou
 ```sh
 $ export KUBECONFIG=/home/user/.secrets/clusters/yavin/auth/kubeconfig
 $ kubectl get nodes
-NAME                                          STATUS   AGE    VERSION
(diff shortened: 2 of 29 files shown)


dghubble:
_sigh_: This causes the Prometheus alert for etcd to break on DigitalOcean clusters.

dghubble:
Fixed by 7de03a12799005a0a7fd1e54d7370459bc81c32d

PR #324, opened 2018-10-19
Disable Kubelet read-only port 10255

dghubble:
* We can finally disable the Kubelet read-only port 10255!
* The lengthy background on getting to this milestone: https://github.com/poseidon/typhoon/issues/322#issuecomment-431073073

Closes: #322

PR #346, opened 2018-11-27
Update pod-checkpointer image to query Kubelet secure API

dghubble:
* Updates pod-checkpointer to prefer the Kubelet secure API (before falling back to the Kubelet read-only API that is disabled on Typhoon clusters since https://github.com/poseidon/typhoon/pull/324)
* Previously, pod-checkpointer checkpointed an initial set of pods during bootstrapping so recovery from power cycling clusters was unaffected, but logs were noisy
* https://github.com/poseidon/terraform-render-bootkube/pull/91

Related:

* https://github.com/kubernetes-incubator/bootkube/pull/1027
* https://github.com/kubernetes-incubator/bootkube/pull/1025


dghubble:
Cluster power cycles nicely and without the error logs.

PR #351, opened 2018-11-29
Add Grafana login information

irontoby:
So simple but I'm embarrassed how long it took me to find the login screen much less the default password!

### DIFF
--- a/docs/addons/grafana.md
+++ b/docs/addons/grafana.md
@@ -12,7 +12,7 @@ Use `kubectl` to authenticate to the apiserver and create a local port-forward t
 kubectl port-forward grafana-POD-ID 8080 -n monitoring
 ```
 
-Visit [127.0.0.1:8080](http://127.0.0.1:8080) to view the bundled dashboards.
+Visit [127.0.0.1:8080](http://127.0.0.1:8080) to view the bundled dashboards. Click the Sign In button in the lower-left corner and use the username/password `admin`/`admin` to create more dashboards, add users, etc.
 
 ![Grafana Capacity Planning](../img/grafana-capacity.png)
 ![Grafana Control Plane](../img/grafana-control-plane.png)


dghubble:
The intent is to provide a read-only view dashboard and let it be visible to anyone with network access (i.e. kubectl port-forward or same LAN).

I'd like to fully disable logins (in this use case, there is no need at all), but Grafana doesn't have a way to disable the admin user.

If you are wanting to manage Grafana directly via accounts (sounds like the case), this is a way but its not a goal. There are upstream Grafana examples more geared for that use case.

dghubble:
For our case, there's a disable_login_form=true we can add to get example manifests closer to the intended read-only dashboard. Though this just removes the form. I don't get the sense Grafana is fully onboard with "we only configuration via Kubernetes".

For your case, correction, this Grafana example isn't suited for managing user accounts. You could currently login to change dashboards, but `basic_auth=false` disables the Grafana user/password system (except `admin`) so probably not what you are wanting.

dghubble:
So this is more of a bug report where the addon isn't configured as intended. #352 improves it.

irontoby:
I don't want to provide public logins; I just want to add more dashboards and the only way to do that after it's deployed is to log in. I understand the read-only intent but the included dashboards are not enough.

At the very least it would be nice to have a note on there that disabling logins is an intentional decision and give the user info on where to go if they want a more fully featured Grafana install or want to add more dashboards.

Even better would be to split up `dashboards.yaml`, and create a `dashboards` directory with individual files and a comment to make it more obvious how to add additional ones (these would then create separate `json` files instead of the current `deployment-dashboard.json`).

Given that the NGINX Ingress Controller is a featured addon for Typhoon, I think including the excellent [NGINX Ingress controller dashboard](https://github.com/kubernetes/ingress-nginx/tree/master/deploy/grafana/dashboards) in particular would be helpful.

If you're interested in making any of these changes, I could submit a pull request for any/all of the above.

dghubble:
Dashboards can be added by editing the configmaps and reloading if you like.

I've been eyeing adding the Nginx dashbords, now that they've worked out some of the bugs. If you're motiviated, you could propose that inline. However, the examples should all be applied via `kubectl apply` only (and without templating), they're setup this way for consumption by other systems.

The addons (not just Grafana) may not be exactly to your liking. They're fairly focused/minimal examples that get validated with Typhoon. Feel free to use upstream manifests or your own if you prefer, there's nothing special about the example ones.

Or phrased another way, I'm not aiming to differentiate with an ecoystem of addons like the enterprisey distros. They're the same apps at the end of the day. Plus Prometheus, Grafana, etc. docs cover their apps and advanced cases way better than I could hope to.

dghubble:
Maybe just a very broad one-liner in docs would suffice ("See the Grafana docs to configure Grafana for your use cases"), though it may just be implied

Issue #392, opened 2019-02-01
Bootstrapping in existing vpc


rverma-nikiai:
An option to create the cluster in existing vpc along with existing external Ips for NAT gateway, or existing NAT gateway.

dghubble:
Existing VPC support was intentionally not added.

On providers with virtual network concepts, each cluster creates its own network (AWS VPC, GCP network, Azure network) that defines IP addressing, subnetting, gateways, routes, subnets, or firewall rules (AWS security groups, Azure security rules). Users may define resources to add or extend the VPC/network (e.g. add security groups), but the network is managed under the purview of the cluster.

When cross-cluster traffic is required, clusters' VPCs/networks should be peered or tunneled explicitly. Design for multiple clouds from day one (e.g. what are you going to do when the other cluster isn't on AWS?). Avoid relying on provider-specific cross-traffic mechanisms (like sharing the same VPC) that lead to lock-in.

A few years ago, customer asks required me to design CoreOS Tectonic clusters to allow existing AWS VPCs. Its a troublesome design.

* Existing VPCs introduce a lot of complexity. Its error-prone. Passing in a pre-existing network/VPC resource creates multiple modes of use, one path creating the VPC, one path for bring-your-own VPC (in various states). An existing VPC would have to be configured to avoid conflicts or collisions with resources created in the module, but those assumptions can't be enforced well. There's also the complexity of build-up - if a partially configured existing VPC were provided, to what degree should missing required resources be added vs assumed.

* Existing VPCs reduce the ease of use and make a module harder to reason about - the module should create a self-contained set of managed, cohesive resources needed for and owned by the cluster. One cluster per VPC. Regardless of cloud.
* Existing VPCs fail the "can we write good documentation" litmus test. I've had to write the docs on it before, and its a mess to explain and worse to support.
* Existing VPC support limits our ability to make cluster virtual network architectural improvements. It changes responsibility for the VPC from managed by the module to a mix that leads to breaking changes, painful migration docs, or upset users in the future.
* Finally, asks for existing VPC only come from AWS users. Take that as you will, but its a design hint. Consider the thought experiment, would we want to extend this design to other platforms? In my mind, its not a good pattern. AWS can't be special.

I'm aware plenty of organizations have historical VPC designs (e.g. everything in one big VPC) that require shared VPCs. But those are out of scope here, this project is able to draw the line, enforce the right practices, and focus on stated goals.

[...]

PR #393, opened 2019-02-01
Existing vpc support

frimik:
Support existing vpc.

By passing in a `vpc_id` and a list of subnet ids in `public_subnets` the choices will be made properly to deploy there instead of creating all the network layer things.

* Optional support for provisioning in an existing vpc (aws/container-linux)
* sanitize the dns_zone variable because if zones come in from other Route53 data sources they will usually have trailing dots.
* TODO: the other stacks
* TODO: support for both public and private subnets
* TODO: support a choice for tying the API and worker NLBs to either public or private subnets.

## Testing

Launched functional clusters in both "modes".

rel: #392

### DIFF
--- a/aws/container-linux/kubernetes/bootkube.tf
+++ b/aws/container-linux/kubernetes/bootkube.tf
@@ -1,9 +1,13 @@
+locals {
+  dns_zone = "${replace(var.dns_zone, "/[.]$/", "")}"
+}
+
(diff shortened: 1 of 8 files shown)


dghubble:
Thanks very much for your proposal, but existing VPC is not a goal. Please see #392

dghubble:
Thanks for the proposal, but I'd rather not go this direction.

PR #394, opened 2019-02-01
aws: support separate dns provider

frimik:
You may want to be able to deploy clusters in AWS accounts which are separate from the accounts where you have your Route53 zone or zones.

* Support a `aws.dns` provider to be supplied. For when you have DNS in one AWS account and the clusters are to be setup in a separate account.

## Testing

Deployed clusters using the same aws providers (accounts) as well as two separate ones.


frimik:
Concept only done for aws/container-linux. For feedback before doing anything else.

dghubble:
Thanks for the proposal, but I'd rather not go this direction.

PR #422, opened 2019-03-12
Configure etcd on Digital Ocean to listen on private IPv4 address only

devurandom:
This is supposed to slightly improve security.

Since we do not yet know the IP address of the machine when rendering the
template in Terraform, we use Container Linux Config's support for CoreOS
Metadata to resolve it during installation.  This requires that we use the
Container Linux Config `etcd` structure instead of manipulating the
`etcd-member.service` directly, because the Container Linux Config transpiler
will only interpolate `{}` statements in that section, but not in `systemd`
units.

Since the Container Linux Config transpiler currently provides no way to set
the `ETCD_SSL_DIR` which `/usr/lib/coreos/etcd-wrapper` will mount into the
container, we still need to override this environment variable in
`etcd-member.service` ourselves.

See-also: https://github.com/coreos/bugs/issues/2565
Signed-off-by: Dennis Schridde <devurandom@gmx.net>

---

## Testing

I deployed a new Kubernetes cluster on Digital Ocean and saw that:
* `/etc/systemd/system/etcd-member.service.d/20-clct-etcd-member.conf` on the target machine was created as expected, with `--listen-peer-urls="https://${COREOS_DIGITALOCEAN_IPV4_PRIVATE_0}:2380"` and similar
* `netstat -nptul` on that machine shows that `etcd` is listening on the private address only
* the Kubernetes cluster was actually starting according to the system journal
* `kubectl --kubeconfig=assets/auth/kubeconfig get all` on the host used to deploy the cluster outputs `service/kubernetes   ClusterIP   10.3.0.1     <none>        443/TCP   ...` and hence confirms that the cluster is healthy.

---

If the chances of this PR getting merged are good, I could port this change to the other cloud providers.

### DIFF
--- a/digital-ocean/container-linux/kubernetes/cl/controller.yaml.tmpl
+++ b/digital-ocean/container-linux/kubernetes/cl/controller.yaml.tmpl
@@ -1,4 +1,22 @@
 ---
+etcd:
+  version: "3.3.12"
+  name: "${etcd_name}"
+  advertise_client_urls: "https://${etcd_domain}:2379"
+  initial_advertise_peer_urls: "https://${etcd_domain}:2380"
+  listen_client_urls: "https://{PRIVATE_IPV4}:2379"
+  listen_peer_urls: "https://{PRIVATE_IPV4}:2380"
+  listen_metrics_urls: "http://{PRIVATE_IPV4}:2381"
+  initial_cluster: "${etcd_initial_cluster}"
+  strict_reconfig_check: true
+  trusted_ca_file: "/etc/ssl/certs/etcd/server-ca.crt"
+  cert_file: "/etc/ssl/certs/etcd/server.crt"
+  key_file: "/etc/ssl/certs/etcd/server.key"
+  client_cert_auth: true
+  peer_trusted_ca_file: "/etc/ssl/certs/etcd/peer-ca.crt"
+  peer_cert_file: "/etc/ssl/certs/etcd/peer.crt"
+  peer_key_file: "/etc/ssl/certs/etcd/peer.key"
+  peer_client_cert_auth: true
 systemd:
   units:
     - name: etcd-member.service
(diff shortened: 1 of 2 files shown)


devurandom:
This has the drawback that the etcd metrics vanish from Prometheus, probably because it scrapes `__meta_kubernetes_node_address_InternalIP` and that is actually set to the external IP address instead of the internal:
```
$ kubectl --kubeconfig=assets/auth/kubeconfig describe node/dev-controller-0 | grep InternalIP
  InternalIP:  157.X.X.X
```

This might be caused by Kubernetes not having a Digital Ocean cloud provider built-in (a mechanism that is deprecated in Kubernetes 1.13) and the Digital Ocean Typhoon config not using the new `external` [Digital Ocean cloud provider](https://github.com/digitalocean/digitalocean-cloud-controller-manager), yet.

dghubble:
Listening on 0.0.0.0 is fine since firewall rules prohibit access and failing that, TLS client auth is required. But if you wanted to change the listen IP, its better to use systemd environment expansion and `coreos-metadata.service` directly so that everything is declared as a systemd unit, networkd unit, etc.

`terraform-provider-ct`'s special `etcd` block and `platform` feature are syntactic sugar for doing the same, but they're less clear, create reliance on that special casing, and introduce the issue of ct now needing to match available etcd flags (as you [noticed](https://github.com/coreos/bugs/issues/2565)). I think its regrettable etcd special cases crept into `ct`. But it doesn't matter here. Favor just using plain systemd.

dghubble:
DO node metadata having the wrong InternalIP is a regression on DO since v1.12.3. Its not related to CCM or cloud-provider (which is unused on every platform). #424

devurandom:
> Listening on 0.0.0.0 is fine since firewall rules prohibit access and failing that, TLS client auth is required. But if you wanted to change the listen IP, its better to use systemd environment expansion and `coreos-metadata.service` directly so that everything is declared as a systemd unit, networkd unit, etc.

How do I make systemd expand environment variables when setting environment variables?  `systemd.exec(5)` specifically says "The $ character has no special meaning."

dghubble:
Variable expansion isn't done when writing out the unit or dropin and they're not eval'd. You'd set the unit to require `coreos-metadata.service` and use variables from `/run/metadata/coreos` in the ExecStart (which does mean you'd override the default ExecStart). `ct` also explains this is what their [etcd block](https://github.com/coreos/container-linux-config-transpiler/blob/master/doc/dynamic-data.md#behind-the-scenes) does, which was carried over to `terraform-provider-ct`.

Overall though, I'm not sure what problem this is solving though. The current setup is concise and doesn't need the metadata services (which can fail). And etcd setups should stay consistent across the 9 platform/OS combos and it doesn't feel worthwhile to go back to having more metadata service dependencies when effort has been made to reduce/eliminate the need (DO kubelet will be the only exception to fix #424 sadly). If you were super set on it, it could also be done with snippets overrides.

dghubble:
Oh, and the other reason I don't want to start relying on `terraform-provider-ct` (and ct's) special etcd-only templating feature is that etcd gets bumped on each release. The corollary to [#2565](https://github.com/coreos/bugs/issues/2565) is that reliance on the syntax sugar means for etcd flag changes, Typhoon would suddenly need to wait on `ct` releases knowing about the flag, re-vendoring that into `terraform-provider-ct`, and requiring everyone to update the plugin, before Typhoon could use the flag. The sugar has that cost.

dghubble:
Hope that explains some of the current reasoning

dghubble:
I'm going to close this one

PR #450, opened 2019-04-11
Nginx ingress addon: remove unused 'health' port in deployment

JordanP:
It has no real purpose and it's not defined in the upstream resource
https://github.com/kubernetes/ingress-nginx/blob/master/deploy/with-rbac.yaml
either.


dghubble:
The health port is used by health checks for the load balancer provided in AWS, Azure, GCP, and DO (if added) clusters. Its also useful for certain bare-metal load balancer setups.

Removal would result in no healthy ingress controller backends.

PR #451, opened 2019-04-11
Fix "google_compute_target_pool.workers: Cannot determine region"

JordanP:
If no region is set at the Google provider level, Terraform fails to
create the google_compute_target_pool.workers resource and complains
with "Cannot determine region: set in this resource, or set provider-level 'region' or 'zone'."

This commit fixes the issue by explicitly setting the region for the
google_compute_target_pool.workers resource.


dghubble:
Looks fine, since the module requires the region as input anyway. Thanks.

The [tutorial](https://typhoon.psdn.io/cl/google-cloud/#provider) will continue to show setting a region for the provider though.

PR #457, opened 2019-04-30
TF 0.12: module source: explicit usage of relative path

JordanP:
With TF 0.11 it is strongly encouraged to use "./" prefix in module
path to indicate that the address is a relative filesystem
path. See [1]

With TF 0.12 this is now required. See [2].

This patch adds the "./" prefix where needed to ensure compatibility
with TF 0.12.

[1] https://www.terraform.io/docs/modules/sources.html#local-paths
[2] https://github.com/hashicorp/terraform/issues/19745


dghubble:
Thanks!

PR #462, opened 2019-05-04
Update etcd from v3.3.12 to v3.3.13 on Container Linux

dghubble:
* Skip updating etcd for Fedora Atomic clusters, now that Fedora Atomic has been deprecated


dghubble:
Logs, metrics, and graphs look reasonable compared with prior version.

PR #472, opened 2019-05-17
Allow Calico networking on Azure and DigitalOcean

dghubble:
* Introduce "calico" as a `networking` option on Azure and DigitalOcean using Calico's new VXLAN support (similar to flannel). Flannel remains the default on these platforms for now.
* Historically, DigitalOcean and Azure only allowed Flannel as the CNI provider, since those platforms don't support IPIP traffic that was previously required for Calico.
* Looking forward, its desireable for Calico to become the default across Typhoon clusters, since it provides NetworkPolicy and a consistent experience
* No changes to AWS, GCP, or bare-metal where Calico remains the default CNI provider. On these platforms, IPIP mode will always be used, since its available and more performant than vxlan


dghubble:
Documenting use of Calico on Azure and DigitalOcean (i.e. VXLAN) mode as experimental. It does work, but my sense from testing this a decent amount is that its pretty early. On DigitalOcean, I've seen Calico restart a decent amount and raise node resource usage. On Azure, I've seen variable bandwidth, but some of this is hard to isolate from the lower reliability/consistency of Azure itself.

Calico (IPIP) remains as stable and reliable as ever on AWS, GCP, and bare-metal.

PR #486, opened 2019-05-29
Migrate from Terraform v0.11 to Terraform v0.12

dghubble:
[Terraform v0.12](https://www.terraform.io/upgrade-guides/0-12.html#upgrades-for-reusable-modules) is a major Terraform release with breaking changes to the HCL language. Language workarounds (e.g. list [type hints](https://www.terraform.io/upgrade-guides/0-12.html#referring-to-list-variables)) have new meanings so it is not possible to provide both v0.11 and v0.12 compatibility at the same time.

* Migrate DigitalOcean module from Terraform v0.11 to v0.12
* Migrate AWS module from Terraform v0.11 to v0.12
* Migrate Google Cloud module from Terraform v0.11 to v0.12
* Migrate Azure module from Terraform v0.11 to v0.12
* Migrate bare-metal module from Terraform v0.11 to v0.12 
* Define `clc_snippets` type constraint map(list(string)) (bare-metal only)
* Use expression syntax instead of interpolated strings, where suggested
* Replace v0.11 bracket type hints with Terraform v0.12 list expressions
* Update tutorials and worker pools documentation
* Define Terraform and plugin version requirements in versions.tf
  * Require digitalocean ~> v1.3 to support Terraform v0.12
  * Require aws ~> 2.7 to support Terraform v0.12
  * Require google ~> 2.5 to support Terraform v0.12
  * Require azurerm ~> 1.27 to support Terraform v0.12
  * Require matchbox ~> 0.3.0 to support Terraform v0.12 (3rd party)
  * Require ct ~> 0.3.2 to support Terraform v0.12 (3rd party)
  * Require terraform-provider-template ~> 2.1
  * Require terraform-provider-null ~> 2.1
  * Require terraform-provider-tls ~> 2.1 (implicit from terraform-render-bootkube)
  * Render terraform-provider-local ~> 1.2 (implicit from terraform-render-bootkube)
* Provide migration instructions from Terraform v0.11 to v0.12
  * Guide on upgrading Terraform for existing resource either in-place or by moving resources to a new Terraform v0.12 config directory one by one

Bug Fixes:

* Avoid unneeded rotations of Regular priority virtual machine scale sets (Azure)
  * Azure only allows `eviction_policy` to be set for Low priority VMs. Supporting Low priority VMs meant when Regular VMs were used, each `terraform apply` rolled workers, to set eviction_policy to null.
  * Terraform v0.12 nullable variables fix the issue and plan does not produce a diff

Closes #483
[...]

PR #488, opened 2019-05-31
Update Kubernetes from v1.14.2 to v1.14.3

dghubble:
* https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#v1143

Related: https://github.com/poseidon/terraform-render-bootkube/pull/137

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@ Notable changes between versions.
 
 ## Latest
 
+## v1.14.3
+
+* Kubernetes [v1.14.3](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#v1143)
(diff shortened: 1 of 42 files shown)


dghubble:
I'll be running Kubernetes v1.14.3 through the forward branch https://github.com/poseidon/typhoon/pull/486, which contains v1.14.3, but also has the upcoming transition to Terraform v0.12. This is done because (1) I've migrated to Terraform v0.12 and (2) master will transition to Terraform v0.12 immediately after the v1.14.3 release.

Migration to Terraform v0.12 was delayed to allow one more release (v1.14.3) with Terraform v0.11 to give folks a Terraform migration window and knowing that v1.14.3 contained a CVE fix (i.e. so delaying migration so as not to discourage patching).

PR #499, opened 2019-06-20
Remove Fedora Atomic documentation

dghubble:
* Typhoon for Fedora Atomic was deprecated in March 2019
* https://typhoon.psdn.io/announce/#march-27-2019


dghubble:
Press heart to pay respects for Fedora Atomic.

Issue #192, opened 2018-04-16
ECR does not work (missing --cloud-provider=aws opt for kubelet)


mabushey:
## Bug
ECR pull does not work

### Environment

* Platform: aws
* OS: CoreOS
* Terraform: v0.11.7
* Plugins: Provider plugin versions
+ provider.aws v1.14.1
+ provider.ct (unversioned)
+ provider.local v1.1.0
+ provider.null v1.0.0
+ provider.template v1.0.0
+ provider.tls v1.1.0

### Problem

Unable to pull images from AWS ECR

Kubelet needs to run with `--cloud-provider=aws` as a CLI argument so that it can get the creds for ECR.

dghubble:
Thanks for the interest and sorry for the answer here. But this is a "won't implement".

Kubelet `--cloud-provider` is not just a simple flag unfortunately. Early in Kubernetes history, some clouds added code to core itself behind the flag. Over time, this became unwieldy and its now widely considered regrettable - there is a moratorium on cloud-provider code and a move toward breaking functionality out of Kubernetes itself (CCM, CSI, flexvolumes). Enabling the AWS cloud provider requires a particular hostname and resource tagging structure, requires elevated IAM roles be granted to nodes, and enables a few anti-patterns I don't wish to support. I was unaware AWS was also using that code path for ECR too. With Typhoon, I aim to provide conformant Kubernetes and offer cloud-provider specific functionality via the newer proper systems as it becomes available and time permitting. I recognize some folks would like cloud-provider despite its hackiness.

I'm not picking on AWS. I'd say the same for any cloud that baked their features inside Kubernetes. It just happens this comes up with AWS a bunch.

[...]

PR #513, opened 2019-07-18
Adds variable for cloud provider in bootkube

vaibhavkhurana2018:

The variable for cloud provider was not there on bootkube. This is resulting in blank value of cloud provider for bootkube because of the default value given in: https://github.com/poseidon/terraform-render-bootkube/blob/master/variables.tf#L21

This PR just gives the user flexibility to add the same. 

## Testing

Have tested the same fro AWS and it is working fine. The kubelet gets the cloud provider as aws.

### DIFF
--- a/aws/container-linux/kubernetes/bootkube.tf
+++ b/aws/container-linux/kubernetes/bootkube.tf
@@ -13,5 +13,6 @@ module "bootkube" {
   cluster_domain_suffix = var.cluster_domain_suffix
   enable_reporting      = var.enable_reporting
   enable_aggregation    = var.enable_aggregation
+  cloud_provider        = var.cloud_provider
 }
 
--- a/aws/container-linux/kubernetes/variables.tf
+++ b/aws/container-linux/kubernetes/variables.tf
@@ -154,3 +154,8 @@ variable "enable_aggregation" {
   default = "false"
 }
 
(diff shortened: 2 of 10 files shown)


dghubble:
Hello. Its intentional that `--cloud-provider` is not used. I've written about this [before](https://github.com/poseidon/typhoon/issues/192#issuecomment-381814562).

> Kubelet --cloud-provider is not just a simple flag unfortunately. Early in Kubernetes history, some clouds added code to core itself behind the flag. Over time, this became unwieldy and its now widely considered regrettable - there is a moratorium on cloud-provider code and a move toward breaking functionality out of Kubernetes itself (CCM, CSI, flexvolumes). Enabling the AWS cloud provider requires a particular hostname and resource tagging structure, requires elevated IAM roles be granted to nodes, and enables a few anti-patterns I don't wish to support. Typhoon aims to provide conformant Kubernetes and only offer cloud-provider specific functionality via proper systems as they become available.

So yes, the flag shows up yes, but none of the old functionality you may be seeking is present (by design, nodes receive no IAM roles). Generally, consider what underlying goal you're trying to achieve. For example, if its storage, look at AWS's CSI driver. If its service Type=LoadBalancer, know that there are better/cheaper ingress mechanisms (see docs).

PR #520, opened 2019-07-24
Fix broken links

DanyC97:

Fix the broken links in the fedora-coreos's bare-metal.md file


DanyC97:
ignore my PR, i just saw that via the official doc it works, just not if you look at the md file individually

PR #526, opened 2019-08-06
Support adding data disks to workers in Azure.

kouk:
When using local storage, or something like [StorageOS](https://storageos.com),
there might be a need to add one or more data disks to the workers. With this patch
the `azure/container-linux/kubernetes/workers/` module supports a new variable
`data_disks` which allows for one or more volumes to be mounted in the worker
like this:

    module "azure-monitoring-low-priority-pool" {
      source = "git::https://github.com/poseidon/typhoon//azure/container-linux/kubernetes/workers?ref=v1.15.1"
      ...

      data_disks = [{
        disk_size  = 500
        disk_type  = "StandardSSD_LRS"
        mount_path = "/media/volume"
      }]
    }

The list can also be passed via the top level azure module as `worker_data_disks`.

## Testing

I've created a functional clusters in Azure, with zero, one or multiple entries in `worker_data_disks` and verified disks are properly mounted.

### DIFF
--- a/azure/container-linux/kubernetes/variables.tf
+++ b/azure/container-linux/kubernetes/variables.tf
@@ -58,6 +58,23 @@ variable "disk_size" {
   description = "Size of the disk in GB"
 }
 
+variable "worker_data_disks" {
+  type        = list(object({
+    disk_size      = number
+    disk_type      = string
+    mount_path     = string
+  }))
+  default     = []
+  description = <<EOD
+A list of maps describing data disks to add to the worker VMs. Disk size in GB. E.g.:
+[{
+  disk_size  = 50
+  disk_type  = "StandardSSD_LRS"
+  mount_path = "/media/foo"
+}]
+EOD
+}
+
 variable "worker_priority" {
   type        = string
   default     = "Regular"
--- a/azure/container-linux/kubernetes/workers.tf
+++ b/azure/container-linux/kubernetes/workers.tf
@@ -20,5 +20,5 @@ module "workers" {
   service_cidr          = var.service_cidr
   cluster_domain_suffix = var.cluster_domain_suffix
   clc_snippets          = var.worker_clc_snippets
+  data_disks            = var.worker_data_disks
 }
-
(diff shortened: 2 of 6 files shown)


dghubble:
Hey Konstantinos, thanks for sharing what you're using. However, this isn't the preferred direction for adding extra disks to worker instances. Instead, try to define managed disks and disk attachments in Terraform and use CLC snippets to inject just your custom mount units. Typhoon modules don't need to be concerned with this directly.

The module is strictly focused on Kubernetes clusters. For example, Typhoon doesn't take as inputs a set of custom load balancers, custom firewall rules, or custom disk configuration. Instead, the underlying goals are accomplished through extension and wiring together module outputs with other Terraform resources. Like programming, extending and composing well-scoped blocks is better than making a really big block that does everything. Have a look at how custom firewall rules are [composed for inspiration](https://typhoon.psdn.io/architecture/azure/#firewalls).

A better proposal would be outputting the worker instance ids you need.

kouk:
I understand, in fact I've been doing what you suggested for custom
firewall rules in my deployment. However it seems to me that it is not
actually possible to achieve this goal in this way, at least on Azure if we
use the virtual machine scale set resource. Even without taking auto
scaling into account, I don't see how terraform allows you to retrieve the
IDs of the VMs in the scaleset, in order to attach a disk. Perhaps in other
cloud providers it is possible but it seems to me that azure doesn't treat
vms in a scaleset in the same way.

dghubble:
True, dealing with GCP instance groups, AWS auto-scaling groups, or Azure VMSS the expectation would be to extend via the group resource (rather than instance resource). Sounds like you have a feature request for [azurerm](https://github.com/terraform-providers/terraform-provider-azurerm) to make VMSS and extra disks composable.

kouk:
Do you have any example handy of how this would work in AWS or GCP? Happy
to submit a feature request to Azure of course..

dghubble:
I don't, its not a case I have. AWS/GCP may well have their own inadequacies. Yes please, discuss VMSS and disks with Azure, I'm waiting on a variety of features in that department.

I'm going to close this out. Adding extra disks to workers shouldn't be in-scope for the module directly. When azure provides extensible mechanisms, the module can output the VMSS to allow this kind of feature to be cleanly done on the user's end. But for now, I'd rather not have the special cases.

dghubble:
Yeah, even GCP's instance templates only have inline disk blocks. This may just not be a very common need. But in the mean time, you've got this patch you can carry for however long you need it. You might be better off using Azure CSI implementations for storage in general on Azure, but I'm sure you've got reasons for building your own atop disks.

PR #536, opened 2019-09-06
 Migrate from self-hosted to static pod control plane

dghubble:
* Run a kube-apiserver, kube-scheduler, and kube-controller-manager static pod on each controller node. Previously, kube-apiserver was self-hosted as a DaemonSet across controllers and kube-scheduler
and kube-controller-manager were a Deployment (with 2 or controller_count many replicas).
* Remove bootkube bootstrap and pivot to self-hosted
* Remove pod-checkpointer manifests (no longer needed)

Details: https://github.com/poseidon/terraform-render-bootkube/pull/148

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ Notable changes between versions.
 
 ## Latest
 
+* Migrate control plane from self-hosted to static pods ([#536](https://github.com/poseidon/typhoon/pull/536))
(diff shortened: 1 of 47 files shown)


dghubble:
I've been running the new Typhoon static control plane for a few weeks and have migrated to using it for my clusters across all Typhoon platforms. No change to conformance status. Going forward into v1.16, I think this is the architecture of choice.

swade1987:
Nice work @dghubble 👍

PR #539, opened 2019-09-10
Change Azure default controller_type and worker_type

dghubble:
* Change default `controller_type` to `Standard_B2s`. The `B2s` is cheaper by $17/month and provides 2 vCPU, 4GB RAM (vs 1 vCPU, 3.5GB RAM)
* Change default `worker_type` to `Standard_DS1_v2`. `F1` was the previous generation. The `DS1_v2` is newer, similar cost, more memory, and still supports Low Priority mode, if desired


dghubble:
Over the last week, I've observed this reduces etcd fsync latency on Azure as well, independent of the recent etcd v3.4 update. Azure SSD disk latency is typically worse than what's observed on AWS and Google Cloud (~10ms), but this brings it closer to them (from ~120ms to ~50ms). Its hard to be absolute about this, as Azure's performance isolation is unknown - its possible I've landed on quiet hosts, but I think not, Azure has been in the ~120ms ballpark for months.

Issue #429, opened 2019-03-15
Allow node-labels and taints on platforms with heterogeneous nodes


dghubble:
AWS, Azure, and Google platforms have internal modules for creating worker pools of different nodes. Its reasonable to allow specifying node-labels (and probably taints) so workloads can distinguish between the workers they'd prefer to run on (e.g. app X needs an on-demand node, app Y needs large local disks, etc).

Bare-Metal has heterogeneous hardware and could use custom node-labels and taints. These need to be sepcified per node, similar to snippets. Configuration becomes too cumbersome with yet more ordered lists or mappings.

The upcoming Terraform v0.12 and language changes will help address this, but the switchover will be tricky and there will be a while before Terraform v0.12 is assumed.

There is no need for custom labels and taints among homogeneous workers.

PR #550, opened 2019-09-24
Add node_labels variable to aws/container-linux/kubernetes/

valer-cara:

* Add `node_labels` variable to kubernetes module (currently only for aws/container-linux)

## Testing

Launched a new cluster with extra labels that were successfully applied on the worker nodes.

----

This is a quick fix I needed for dedicated worker pools and nodeSelector. I guess that for this to be merged it'd require uniformity across all cloud/distro variants.

I'm submitting it just to see if this would be an acceptable feature. If so, I can add it to the rest of the platforms/distros.

Thanks!

### DIFF
--- a/aws/container-linux/kubernetes/workers/cl/worker.yaml.tmpl
+++ b/aws/container-linux/kubernetes/workers/cl/worker.yaml.tmpl
@@ -61,6 +61,9 @@ systemd:
           --lock-file=/var/run/lock/kubelet.lock \
           --network-plugin=cni \
           --node-labels=node.kubernetes.io/node \
+          %{ for label in split(",", node_labels) }
+          --node-labels=${label} \
+          %{ endfor ~}
           --pod-manifest-path=/etc/kubernetes/manifests \
           --read-only-port=0 \
           --volume-plugin-dir=/var/lib/kubelet/volumeplugins
--- a/aws/container-linux/kubernetes/workers/workers.tf
+++ b/aws/container-linux/kubernetes/workers/workers.tf
@@ -86,6 +86,7 @@ data "template_file" "worker-config" {
     cluster_dns_service_ip = cidrhost(var.service_cidr, 10)
     cluster_domain_suffix  = var.cluster_domain_suffix
     cgroup_driver          = local.flavor == "flatcar" && local.channel == "edge" ? "systemd" : "cgroupfs"
+    node_labels            = join(",", var.node_labels)
   }
 }
 
--- a/aws/container-linux/kubernetes/variables.tf
+++ b/aws/container-linux/kubernetes/variables.tf
@@ -154,3 +154,9 @@ variable "enable_aggregation" {
   default = "false"
 }
 
+variable "worker_node_labels" {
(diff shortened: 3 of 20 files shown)


rochacon:
👍 I think this is useful, likewise, a similar feature to support taints (e.g. `--register-with-taints`) would be nice.

aws/container-linux/kubernetes/workers/cl/worker.yaml.tmpl
```diff
@@ -61,6 +61,9 @@ systemd:
           --lock-file=/var/run/lock/kubelet.lock \
           --network-plugin=cni \
           --node-labels=node.kubernetes.io/node \
+          %{ for label in split(";", node_labels) }
```
rochacon: Since `var.node_labels` is of type `list`, can't we get rid of the `join/split` here?


(Sorry if useless

valer-cara: that was my initial take too, but tf only takes strings as template vars.
(https://github.com/hashicorp/terraform/issues/9368 , etc..)

valer-cara:
yup, also had taints in mind, similar pattern; that would work as a separate PR i guess (if acceptable)

JordanP:
I think this is acceptable becauce it's already been reported and supported here: https://github.com/poseidon/typhoon/issues/429

I, too, second this feature.

valer-cara:
There's one more thing slightly related. While typhoon aims to keep things simple, my specific usecase requires cluster autoscaling for some one-time workload happening only on initialization.

In order to fully support CA, i'd also need to be able to add tags to the workers autoscaling group (it uses those `k8s.io/cluster-autoscaler/node-template*` tags to figure out some details). Would that be ok as part of this PR or at all?

dghubble:
Yes, I'll take a `node-labels` PR per #429. Please set them for platforms with worker pools only (AWS, GCP, Azure), leaving out bare-metal and DO which have separate considerations. The inner worker modules can have a `node_labels` variable. The kubernetes module should use `worker_node_labels` (or possibly don't provide it yet since this primarily concerns worker pools).

No changes to tags.

aws/container-linux/kubernetes/workers/workers.tf
```diff
@@ -86,6 +86,7 @@ data "template_file" "worker-config" {
     cluster_dns_service_ip = cidrhost(var.service_cidr, 10)
     cluster_domain_suffix  = var.cluster_domain_suffix
     cgroup_driver          = local.flavor == "flatcar" && local.channel == "edge" ? "systemd" : "cgroupfs"
+    node_labels            = join(";", var.node_labels)
```
dghubble: Let's at least use a "," separator so this is less weird.

valer-cara:
@dghubble - I've made the requested changes. Added node labels to aws (cl + fedora), gcp, azure.

I've only tested the aws/cl setup so far.

dghubble:
Looks good. It'll need updates to docs and tutorials, but its probably simpler if I just add that.

dghubble:
Thanks :+1:

PR #569, opened 2019-10-07
Support worker managed identity in Azure

kouk:
Add a variable `worker_identity` to optionally configure system managed
identity for the worker vms in Azure.

## Testing

Using this in production like this:

    resource "azurerm_role_assignment" "cluster-role" {
      name                 = "${random_uuid.cluster-role-name.result}"
      scope                = "${module.typhoon_cluster.resource_group_id}"
      role_definition_name = "Contributor"
      principal_id         = "${module.typhoon_cluster.worker_principal_id}"
    }

See also: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview

### DIFF
--- a/azure/container-linux/kubernetes/outputs.tf
+++ b/azure/container-linux/kubernetes/outputs.tf
@@ -19,6 +19,10 @@ output "resource_group_name" {
   value = azurerm_resource_group.cluster.name
 }
 
+output "resource_group_id" {
+  value = azurerm_resource_group.cluster.id
+}
+
 output "subnet_id" {
   value = azurerm_subnet.worker.id
 }
@@ -29,6 +33,11 @@ output "security_group_id" {
 
 output "kubeconfig" {
   value = module.bootstrap.kubeconfig-kubelet
+}
+
+output "worker_principal_id" {
+  description = "The identity principal for the worker pool VMs"
+  value       = module.workers.principal_id
 }
 
 # Outputs for custom firewalling
--- a/azure/container-linux/kubernetes/variables.tf
+++ b/azure/container-linux/kubernetes/variables.tf
@@ -46,6 +46,12 @@ variable "worker_type" {
   default     = "Standard_DS1_v2"
 }
 
+variable "worker_identity" {
+  description = "The type of Identity management to specify for the worker VMs. Options are None (default) or SystemAssigned"
+  type = string
+  default = "None"
+}
+
 variable "os_image" {
   type        = string
   default     = "coreos-stable"
--- a/azure/container-linux/kubernetes/workers.tf
+++ b/azure/container-linux/kubernetes/workers.tf
@@ -21,5 +21,6 @@ module "workers" {
   cluster_domain_suffix = var.cluster_domain_suffix
   clc_snippets          = var.worker_clc_snippets
   node_labels           = var.worker_node_labels
+  identity              = var.worker_identity
 }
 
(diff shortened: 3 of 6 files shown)


dghubble:
Cluster nodes are not given account IAM / identity permissions on any cloud platform (e.g. AWS, GCP, Azure). See [security properties](https://typhoon.psdn.io/topics/security/). Container workloads should make use of Kubernetes Secrets or other mechanisms (e.g. Vault) to mount credentials into specific applications. Plenty of cloud service offerings prefer you associating identities with underlying VMs and locked into their cloud, but permitting it isn't a goal here.

If that doesn't match your needs I understand. Probably the best thing is using this as a base and maintaining the modifications you require.

kouk:
hmm, ok. Basically what I'm trying to do is configure the AzureDisk CSI plugin, and currently I'm using this CLC snippet:
```
# custom-files
storage:
  files:
    - path: /etc/kubernetes/azure.json
      filesystem: root
      contents:
        inline: |
          {
              "cloud":"AzurePublicCloud",
              "tenantId": "${tenant_id}",
              "subscriptionId": "${subscription_id}",
              "resourceGroup": "${resource_group}",
              "location": "${location}",
              "subnetName": "worker",
              "vmType": "vmss",
              "securityGroupName": "${cluster_name}-worker",
              "vnetName": "${cluster_name}",
              "vnetResourceGroup": "${resource_group}",
              "cloudProviderBackoff": false,
              "useManagedIdentityExtension": true,
              "useInstanceMetadata": true
          }
      mode: 0644
```
The `useManagedIdentityExtension` option requires the VM to have a managed identity, hence this PR. I'll try and figure out another way, but in the meantime consider a documentation improvement request to add some "storage how-to"  :)

kouk:
@dghubble i'm trying to provision a service principal beforehand via terraform (so I can add it's password to `azure.json`). However to be able to assign it a Contributor role I need access to `azurerm_resource_group.cluster.id` from typhoon. Is the resource group considered an implementation detail?

dghubble:
I'm assuming you're trying to use [azure-csi-driver](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/docs/driver-parameters.md). From their driver requirements doc, it looks like they want the resource group name, which is already in outputs.

Just be aware adding custom components within the cluster's resource group can impede cluster deletion, be sure to delete custom components along with the cluster.

kouk:
@dghubble  what I need the ID for is configuring the [scope|https://www.terraform.io/docs/providers/azurerm/r/role_assignment.html#scope] of the role that I assign to the service principal for azure-csi-driver. For example I have this:
```
resource "azurerm_role_assignment" "cluster-role" {
  name                 = "${random_uuid.cluster-role-name.result}"
  scope                = "${module.typhoon_cluster.resource_group_id}"
  role_definition_name = "Contributor"
  principal_id         = "${azuread_service_principal.cluster_sp.object_id}"
}
```

The above example is straight from my config, however I am forking typhoon in order to add the `resource_group_id` output. Currently this is the only reason I keep a fork, so would be great if we could add the resource group id.

dghubble:
I'd accept a PR that just adds an output for the resource group id then. Since the resource group name is ok to output. Just haven't seen cases using the id, since Azure typically references the resource group by name.

kouk:
@dghubble great, here it is: https://github.com/poseidon/typhoon/pull/577

PR #594, opened 2019-11-30
DigitalOcean: Fix outputs

aknuds1:
* Fix list output variables for DigitalOcean

## Testing
I've tested by creating a cluster with Terraform 0.12.16. Referencing variables such as `workers_ipv4` would fail since it was a 1-element tuple containing a list. After my fix it works AFAICT, at least it parses without problems.


dghubble:
The enclosing list was originally required in Terraform v0.11 syntax. With Terraform v0.12, these became a list of list, which is not nice to consume. Thanks!

PR #654, opened 2020-02-26
Use a route table with separate (rather than inline) routes

dghubble:
* Allow users to extend the route table using a data reference and adding route resources (e.g. unusual peering setups)
* Note: Internally connecting AWS clusters can reduce cross-cloud flexibility and inhibits blue-green cluster patterns. It is not recommended


meabed:
Thank for this :) when it would be merged?

meabed:
Awesome Thank you!

PR #700, opened 2020-04-11
Allow target group health check to pass

medined:
When I created an AWS Fedora CoreOS cluster using Typhoon, the instances in the worker target group did not pass the health checks. On each worker, I manually updated the healthz port and added a bind address, then ran "systemctl daemon-reload" and "systemctl restart kubelet". Within minutes, the health checks were passing and my ingress controller was letting requests pass into services.

### DIFF
--- a/aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
+++ b/aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
@@ -61,7 +61,8 @@ systemd:
           --cluster_domain=${cluster_domain_suffix} \
           --cni-conf-dir=/etc/kubernetes/cni/net.d \
           --exit-on-lock-contention \
-          --healthz-port=0 \
+          --healthz-port=10254 \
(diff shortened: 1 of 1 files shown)


dghubble:
The worker target groups for http and https are for Ingress. If you don't install an ingress controller, that's perfectly fine, those health checks won't be met, but you don't need to worry about that since you're not using ingress anyway. If you do choose to deploy an ingress controller satisfying the 10254 health check (e.g. nginx-ingress addon), it ensures traffic only goes to nodes with an ingress controller.

Satisfying 10254 is done by Ingress Controllers and is optional.

medined:
I now understand that the nginx-ingress addon, when installed, responds to the health check. Does the nginx-ingress-controller Deployment (addons/nginx-ingress/aws/deployment.yaml) need to reference the service account? I am seeing the following message: The cluster seems to be running with a restrictive Authorization mode and the Ingress controller does not have the required permissions to operate normally.

PR #726, opened 2020-05-12
Update SELinux to Permissive

vamshisiddarth:
Hello @dghubble. This PR is w.r.t the issue where we are not able to ship the logs from Worker Nodes using services such as FluentD, Datadog. The SELinux "enforcing" label was blocking these services to access `/var/log` location. 

I understand the importance of SELinux to be existent on the instances to block apps from publishing confidential data. However, setting the selinux label to permissive will help everyone who is using the log shipping agents installed on the Typhoon Kubernetes Cluster until there is right support provided for Kubelet service from selinux.

Also, it is recommended by Kubernetes to set the selinux to either "permissive" or "disabled" label to enable kubernetes to overcome the issues in such scenarios as suggested [here](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#k8s_install-1).

I have also enabled sysconfig for docker in this PR. This will allow us to deploy Elasticsearch on our Kubernetes cluster without the error `Exception
java.lang.RuntimeException: max file descriptors [65535] for elasticsearch process likely too low, increase to at least [65536]`.

We have been running our Production environment on the Kubernetes cluster built using the Typhoon module and thought these things bundled in the module will be fruitful to many others looking to deploy services similar to us on kubernetes.

Please let me know your thoughts on this and do let me know if you would like these to be implemented in any other way. 

Looking forward for your reply 😄

### DIFF
--- a/aws/fedora-coreos/kubernetes/fcc/controller.yaml
+++ b/aws/fedora-coreos/kubernetes/fcc/controller.yaml
@@ -207,6 +207,22 @@ storage:
           ETCD_PEER_CERT_FILE=/etc/ssl/certs/etcd/peer.crt
           ETCD_PEER_KEY_FILE=/etc/ssl/certs/etcd/peer.key
           ETCD_PEER_CLIENT_CERT_AUTH=true
+    - path: /etc/selinux/config
+        mode: 0644
+        overwrite: true
+        contents:
+          inline: |
+            # This file controls the state of SELinux on the system.
+            # SELINUX= can take one of these three values:
+            #     disabled - SELinux security policy is enforced.
+            #     permissive - SELinux prints warnings instead of disabled.
+            #     disabled - No SELinux policy is loaded.
+            SELINUX=permissive
+            # SELINUXTYPE= can take one of these three values:
+            #     targeted - Targeted processes are protected,
+            #     minimum - Modification of targeted policy. Only selected processes are protected.
+            #     mls - Multi Level Security protection.
+            SELINUXTYPE=targeted
 passwd:
   users:
     - name: core
--- a/aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
+++ b/aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
@@ -112,6 +112,35 @@ storage:
           DefaultCPUAccounting=yes
           DefaultMemoryAccounting=yes
           DefaultBlockIOAccounting=yes
+    - path: /etc/sysconfig/docker
+      mode: 0644
+      overwrite: true
+      contents:
+        inline: |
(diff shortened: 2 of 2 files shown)


dghubble:
Typhoon for Fedora CoreOS will continue to use SELinux enforcing mode. It is a key security property.

Pods run container processes with SELinux `container_t` and a random MCS pair. For unique workloads, you may adjust the [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#assign-selinux-labels-to-a-container) of the pod. If you absolutely must, run special workloads as privileged for `spc_t`.

Security can require discipline. Even more so if you run components that access host data (like log aggregators for vendors). Make security-conscious choices in what you run (e.g. if a project tells you to first disable SELinux, reject that project), take the time to learn about SELinux to confine special workloads, or hire an expert. Organizations that drop enforcement usually lose it permanently (e.g. either understand 1 vendor product now or struggle to reign in a tangled lax ecosystem years from now). For the sake of your user's and your own organization, I'd suggest committing to remaining enforcing. If you still choose to drop enforcing, I'll ask that you do it privately on your own.

The linked kubeadm tutorial is incorrect (Kubelet does integrate with SELinux, plus the older CoreOS Container Linux (EOL) they're referring to ran permissive) and doesn't amount to a recommendation. Even if a Kubernetes community member made some statement to that effect, it wouldn't hold weight here.

vamshisiddarth:
`The linked kubeadm tutorial is incorrect (Kubelet does integrate with SELinux, plus the older CoreOS Container Linux (EOL) they're referring to ran permissive) and doesn't amount to a recommendation. Even if a Kubernetes community member made some statement to that effect, it wouldn't hold weight here.`
I didn't actually realize that. This will be something new I would be taking from here.

I will try out `spc_t` security context and will let you know how it goes. Thanks for the prompt response 😃

dghubble:
Glad to hear it, I hope choosing contexts for your workloads goes well.

vamshisiddarth:
@dghubble I now have the good news. I tried the below security context for my datadog agent and it is able to ship all the logs magically 😄 

```
securityContext:
    seLinuxOptions:
      user: system_u
      role: system_r
      type: spc_t
      level: s0
```
I have logged into the node and checked `ps -efZ | grep kubelet` which gave me this output(`system_u:system_r:container_runtime_t:s0`). I then referred [this link](https://jaosorior.dev/2019/selinux-and-kubernetes/) and created the above security context and deployed the agent. Please do let me know if I have taken any wrong approach as I'm still new to this set of configurations. :)

If you don't mind, I wanted to suggest to add this piece of info in your website in the addon section so that anyone new to this can make these changes without exploring everything else. 

And, thanks a lot for pointing me in the right direction. Appreciate the thorough explanation on the security part.

PR #780, opened 2020-07-01
Add default tags for AWS resources

kivio:

* Add possibility to add default tags to all tag aware AWS resources in container linux

It's very useful from AWS perspective - you can use it to estimate costs for example. 

## Testing

You need to apply it into aws environment with module source part to: typhoon//aws/container-linux/kubernetes
All resources like load balancer, vpc, subnets, machines, target groups, etc should be tagged with tags you provide in `default_tags` parameter of module.

### DIFF
--- a/aws/container-linux/kubernetes/controllers.tf
+++ b/aws/container-linux/kubernetes/controllers.tf
@@ -17,9 +17,9 @@ resource "aws_route53_record" "etcds" {
 resource "aws_instance" "controllers" {
   count = var.controller_count
 
-  tags = {
+  tags = merge(var.default_tags, {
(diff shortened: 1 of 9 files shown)


dghubble:
I've commented before, but think the AWS "tag every resource" pattern is a poor grouping mechanism compared to designs other clouds offer (projects, resource groups, etc).

I'm ok with AWS tagging resources with `"Name" = var.cluster_name` to accommodate some usage. Cluster names are unique and should be adequate. But I don't want to expose a new AWS-specific variable to plumb tags around or encourage their use too much. Also, FCOS and Flatcar AWS should match in this respect.

PR #788, opened 2020-07-25
Fix 404s

eldondev:
Found a few 404's in the docs, hoping this fixes them. There are some remaining 404's around atomic from the announce page, but it seemed a bit revisionist to remove those, and they don't seem core to navigating the site. I expect these updates to work, but to be honest I haven't run them through the GH-wringer to test them.


dghubble:
Thanks

PR #810, opened 2020-08-20
Fix kubelet cert symlinks in Fedora CoreOS

beyondbill:
`kubelet` container has `/etc/ssl/certs` mounted from the host. `/etc/ssl/certs` is a symlink. 
 In Fedora CoreOS, the symlink points to `/etc/pki/tls/certs` rather than `/usr/share/ca-certificates` in Container Linux.

```
[root@ip-10-20-151-215 scoop-ops]# ls -al /etc/ssl/certs
lrwxrwxrwx. 1 root root 16 Aug 20 02:22 /etc/ssl/certs -> ../pki/tls/certs
```

```
[root@ip-10-20-151-215 scoop-ops]# ls -al /etc/pki/tls/certs
total 0
drwxr-xr-x. 2 root root  54 Aug 20 02:21 .
drwxr-xr-x. 5 root root 104 Aug 20 02:21 ..
lrwxrwxrwx. 1 root root  49 Aug 20 02:21 ca-bundle.crt -> /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
lrwxrwxrwx. 1 root root  55 Aug 20 02:21 ca-bundle.trust.crt -> /etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt
```

In addtion, both certs under `/etc/pki/tls/certs` are symlinks pointing to files under `/etc/pki/ca-trust/extracted`, which needs to be mounted to `kubelet` as well.

### DIFF
--- a/aws/fedora-coreos/kubernetes/fcc/controller.yaml
+++ b/aws/fedora-coreos/kubernetes/fcc/controller.yaml
@@ -74,7 +74,8 @@ systemd:
           --volume /run:/run \
           --volume /sys/fs/cgroup:/sys/fs/cgroup:ro \
           --volume /sys/fs/cgroup/systemd:/sys/fs/cgroup/systemd \
-          --volume /etc/pki/tls/certs:/usr/share/ca-certificates:ro \
+          --volume /etc/pki/tls/certs:/etc/pki/tls/certs:ro \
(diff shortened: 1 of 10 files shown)


dghubble:
Is there a problem you're trying to solve?

PR #835, opened 2020-09-16
Allow for per node install disk in Terraform Installer

carpenike:
I have different disk types on my nodes (nvme disks on one, sda on others).

This change moves the `install_disk` parameter from a global variable into the node specific list.

## Testing

Deployed with my own cluster.

### DIFF
--- a/bare-metal/container-linux/kubernetes/profiles.tf
+++ b/bare-metal/container-linux/kubernetes/profiles.tf
@@ -130,6 +130,10 @@ resource "matchbox_profile" "cached-flatcar-linux-install" {
     var.kernel_args,
   ])
 
+  vars = {
+    install_disk = concat(var.controllers.*.install_disk, var.workers.*.install_disk)[count.index]
+  }
+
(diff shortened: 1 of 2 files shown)


dghubble:
Terraform object passing for node's `name`, `mac`, and `domain` is not ideal, but at least those are required. For optional parameters, I don't want to have to enumerate those per node, its too cumbersome. For now, I'd rather not support the feature.

You could explore using a map like some of the other fields. Also, feature additions go to both OSes.

carpenike:
OK that's fair. Seems like any node specific settings / options that aren't global should be kept together but I understand. I'll close the PR and run with my fork.

dghubble:
Yes, but then existing users have to now duplicate their disk everywhere (until Terraform supports optional fields in objects with defaulting). Other sparsely used fields use a map to avoid this.

Similar to how snippets are accepted https://typhoon.psdn.io/advanced/customization/ ("On Bare-Metal....")

carpenike:
Ah, yes it would be a breaking change.

PR #888, opened 2020-11-25
 coreos.live.rootfs_url to support new separated initramfs for PXE booting

cta101:
Fedora CoreOS PXE initramfs image has been separated into two images: initramfs and rootfs. Both of them needed for PXE boot which brakes the current setup.  The new rootfs artefact can be specified as either a second element in the initrd array or as a kernel argument `coreos.live.rootfs_url=`. 
More information [here](https://github.com/coreos/fedora-coreos-tracker/issues/390#issuecomment-659791267) 

This PR contains the latter. I didn't see that one would load faster than the other in my environment. In the Fedora CoreOS [PXE guide](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/) noted that the kernel argument `coreos.live.rootfs_url` requires at least 2GiB RAM while the initrd option requires at least 3GiB RAM.


dghubble:
Fedora CoreOS profiles in Matchbox were updated, for rootfs image becoming available on stable on Oct 6th, but not here. The PR should match https://github.com/poseidon/matchbox/commit/da0df01763693b6391b3f6f9b7a267818075cef8#diff-4541f7b7c174f6ae6270135942c1c65ed9e09ebe81239709f5a9fb34e858ddcf

Caveat is that everyone must use a sufficiently new FCOS stable version (32.20200923.3.0+) if they aren't already.

dghubble:
Thanks for the reminder and testing out approaches

dghubble:
Superceded by #889

PR #913, opened 2020-12-16
Support for Terraform v0.14.0 

vamshisiddarth:
- [x] Support for Terraform v0.14.0 and higher.


dghubble:
We've already had a change for this but it was reverted until Terraform v0.14 fixes various issues with sentitive marking gone awry. I don't recommend you update yet. https://github.com/poseidon/typhoon/pull/911

PR #920, opened 2020-12-26
create kubeconfig file alongside the cluster

Andrei-Paul:
Made obtaining the kubeconfig file easier/clearer for beginners (I needed this 😅).

* Added kubeconfig file creation along side the cluster definition using https://typhoon.psdn.io/#example as a model.
* Reworded the previous "first encounter" sentence for kubeconfig to point to the Cluster section.

### DIFF
--- a/docs/flatcar-linux/bare-metal.md
+++ b/docs/flatcar-linux/bare-metal.md
@@ -188,6 +188,12 @@ module "mercury" {
   # set to http only if you cannot chainload to iPXE firmware with https support
   # download_protocol = "http"
 }
+
+# Obtain cluster kubeconfig
+resource "local_file" "kubeconfig-mercury" {
+  content  = module.mercury.kubeconfig-admin
+  filename = "/home/user/.kube/configs/mercury-config"
+}
 ```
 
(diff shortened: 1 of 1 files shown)


dghubble:
Thanks for your thoughts on presenting this content.

But I prefer the tutorials show cluster creation, and *then* the use of outputs (such as kubeconfig to verify the cluster was created), rather than moving it into the same block. As is, the tutorial builds up newcomers understanding of what each piece does, incentivizes reading, and it's consistent across the tutorials. I recognize a trade-off is folks can't copy paste one section.

Andrei-Paul:
My issue was that I simply did not know what to do with that piece of config.

After _terraform apply; "All done, you're good to go"_ it's was bit of a stretch to consider to go back through the plan / apply process again after adding that piece of config to... some random file? I was not sure it was supposed to be a new tf file, went with a "try.tf" because of similar style and went through init / plan / apply and it worked. Good for me, but that was unlikely to work in the grand scheme of how any/all software works.

Maybe mention "create this in a new file or append it to your previous cluster file (for witch no name+type was suggested either) and the rerun terraform plan/apply/whatever to get your config".

I'm just coming here with the absolute beginner perspective, so other absolute beginners with less lucky guesses will not be discouraged to experiment with typhoon, I didn't mean to overstep. 🙂

PR #945, opened 2021-03-11
update fedora bare-metal bootstrap

ajrpayne:

Added kubelet image environment variable in Fedora bare-metal bootstrap for use with systemd dropins (Flatcar bare-metal already did this.)

### DIFF
--- a/bare-metal/fedora-coreos/kubernetes/fcc/controller.yaml
+++ b/bare-metal/fedora-coreos/kubernetes/fcc/controller.yaml
@@ -120,6 +120,7 @@ systemd:
         Type=oneshot
         RemainAfterExit=true
         WorkingDirectory=/opt/bootstrap
+        Environment=KUBELET_IMAGE=quay.io/poseidon/kubelet:v1.20.4
         ExecStartPre=-/usr/bin/podman rm bootstrap
         ExecStart=/usr/bin/podman run --name bootstrap \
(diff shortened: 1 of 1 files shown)


bare-metal/fedora-coreos/kubernetes/fcc/controller.yaml
```diff
@@ -120,14 +122,15 @@ systemd:
         Type=oneshot
         RemainAfterExit=true
         WorkingDirectory=/opt/bootstrap
+        Environment=KUBELET_IMAGE=quay.io/poseidon/kubelet:v1.20.4
```
dghubble, line 125: I'll take the change to bootstrap.service in isolation, since its just matching what's done for other platforms. Please remove the unrelated indentation and other changes and update the description.

Also, the intention of these isn't to support customizing every container image that's used in the cluster. Only to allow selection from the quay and docker.io variants of images built by Poseidon. pause would require additional maintenance to manage and track those versions upstream to keep this in sync and full airgap isn't a feature I'm willing to promise.

ajrpayne: I think this is resolved.

I guess for now I was keep a separate branch with my changes for air-gap bare-metal installs.

dghubble:
thanks

PR #951, opened 2021-03-18
Set provider ID for AWS nodes

vixus0:
Sets the provider ID for controller and worker nodes on AWS to `aws:///<az>/<instance id>`.

* Use `afterburn.service` on Fedora and `coreos-metadata.service` on Flatcar to write instance metadata to file.
* Read instance metadata file in `kubelet.service`.
* Set `--provider-id` in kubelet arguments.

## Testing

* Provisioned two clusters on AWS, one each of Fedora and Flatcar.
* Checked kubectl describe node had a correct value for ProviderID in the format `aws:///<az>/<instance id>` for both controllers and workers.

## Rationale

Autoscaling can be pretty essential for many cluster operators.
Without a provider ID, it is impossible (as far as I know) to use tools such as [cluster-autoscaler](kubernetes/autoscaler).

I understand this change isn't ideal as it makes use of the instance metadata service only available on AWS EC2 instances.
However, given there is an `aws` subdirectory, we already have some explicit notion of which cloud provider we're running against.

## Alternative options

* Autoscale some other way, but nothing comes readily to mind.
* Set hostname in kubelet to the value of `hostname --fqdn`.
  This would maybe provide the option for running with an external cloud-controller-manager as it will match the node name to the EC2 instance and set `providerID` accordingly.
  But then we might need a way to set `--cloud-provider=external` as well, not sure.

### DIFF
--- a/aws/fedora-coreos/kubernetes/fcc/controller.yaml
+++ b/aws/fedora-coreos/kubernetes/fcc/controller.yaml
@@ -50,9 +50,12 @@ systemd:
       contents: |
         [Unit]
         Description=Kubelet (System Container)
+        Requires=afterburn.service
+        After=afterburn.service
         Wants=rpc-statd.service
         [Service]
         Environment=KUBELET_IMAGE=quay.io/poseidon/kubelet:v1.20.4
+        EnvironmentFile=/run/metadata/afterburn
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/cni/net.d
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/manifests
         ExecStartPre=/bin/mkdir -p /opt/cni/bin
@@ -93,6 +96,7 @@ systemd:
           --network-plugin=cni \
           --node-labels=node.kubernetes.io/controller="true" \
           --pod-manifest-path=/etc/kubernetes/manifests \
+          --provider-id=aws:///$${AFTERBURN_AWS_AVAILABILITY_ZONE}/$${AFTERBURN_AWS_INSTANCE_ID} \
           --read-only-port=0 \
           --register-with-taints=node-role.kubernetes.io/controller=:NoSchedule \
           --rotate-certificates \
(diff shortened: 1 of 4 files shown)


dghubble:
I think the relevant link from the cluster-autoscaler project is [here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md) with their needs. I'm open to setting `provider-id` for other unrelated reasons (on controllers too).

However, this should not need an additional systemd unit. Fedora CoreOS has [afterburn](https://github.com/coreos/afterburn/blob/master/docs/usage/attributes.md) and Flatcar Linux has a [fork](https://github.com/kinvolk/afterburn/) as well. At present, the only platforms that need afterburn enabled are DigitalOcan so you can see them for [examples](https://github.com/poseidon/typhoon/blob/master/digital-ocean/fedora-coreos/kubernetes/fcc/controller.yaml). There is some potential reliability concern since metadata can be a new flake point in bringup.

vixus0:
Yes, the only bits missing from Typhoon that I could see were the following:

> If you're managing your own kubelets, they need to be started with the --provider-id flag. The provider id has the format aws:///<availability-zone>/<instance-id>, e.g. aws:///us-east-1a/i-01234abcdef.

There's also the requirement for adding extra tags to the auto scaling group but that can be done with a `local-exec` provisioner.

I'll look into using afterburn and then I'll update the PR, thanks for the pointer.

vixus0:
After spending some time faffing around with Fedora's recent SSH key issues I've managed to successfully test the changes using afterburn and coreos-metadata.

Testing:

* Provisioned two clusters on AWS, one each of Fedora and Flatcar.
* Checked `kubectl describe node` had a correct value for `ProviderID` in the format `aws:///<az>/<instance id>`.

dghubble:
Rebased and merged as 507c646e8b2bc11e6084b399eb2c047d7a4a5e4e, thanks Anshul

PR #959, opened 2021-03-23
Experiment with Fedora CoreOS containerd

dghubble:
* Fedora CoreOS ships containerd indirectly as a dependency of docker

### DIFF
--- a/aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
+++ b/aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
@@ -3,7 +3,7 @@ variant: fcos
 version: 1.1.0
 systemd:
   units:
-    - name: docker.service
+    - name: containerd.service
       enabled: true
     - name: wait-for-dns.service
       enabled: true
@@ -47,7 +47,7 @@ systemd:
           --volume /sys/fs/cgroup:/sys/fs/cgroup:ro \
           --volume /sys/fs/cgroup/systemd:/sys/fs/cgroup/systemd \
           --volume /var/lib/calico:/var/lib/calico:ro \
-          --volume /var/lib/docker:/var/lib/docker \
+          --volume /var/lib/containerd:/var/lib/containerd:ro \
           --volume /var/lib/kubelet:/var/lib/kubelet:rshared,z \
           --volume /var/log:/var/log \
           --volume /var/run/lock:/var/run/lock:z \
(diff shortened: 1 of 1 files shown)


aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
```diff
@@ -130,6 +132,18 @@ storage:
...
+      overwrite: true
+      contents:
+        inline: |
+          [plugins]
+            [plugins."io.containerd.grpc.v1.cri"]
+              enable_selinux = true
+              [plugins."io.containerd.grpc.v1.cri".cni]
+                bin_dir = "/opt/cni/bin"
+                conf_dir = "/etc/kubernetes/cni/net.d"
+                max_conf_num = 1
```
wernerb, line 145: you have to enable systemd to avoid running cgroupfs with containerd. Consider adding:
```
          [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
            systemdCgroup = true
```

aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
```diff
@@ -3,7 +3,7 @@ variant: fcos
 version: 1.1.0
 systemd:
   units:
-    - name: docker.service
+    - name: containerd.service
       enabled: true
```
wernerb, line 7: For containerd add an override drop-in:
```
ExecStartPre=-/bin/sh -c "for f in /usr/libexec/cni/*; do ln -s \"$f\" /opt/cni/bin/$(basename $f); done"
```
This symlinks in the cni plugins that are installed in `/usr/libexec/cni`

wernerb (commented):
Also consider adding 
```
ExecStartPre=-/bin/sh -c "for f in /usr/libexec/cni/*; do ln -s \"$$f\" /opt/cni/bin/$$(basename $$f); done"
```

wernerb:
A problem that turns up with containerd is a warning Event message on the kubelet `Node` objects. Something related to not being able to GC the image filesystem.

wernerb:
Be aware that some CNIs just are not supported yet because they rely on dockershim directly. the easy solution is like https://github.com/kubernetes/kops/pull/10502/files to just symlink it as part of `kubelet.service` for compatibility.
```
ExecStartPre=-/bin/sh -c "ln -s /run/containerd/containerd.sock /var/run/dockershim.sock"
```
I have confirmed that aws-vpc-cni-k8s works with this until https://github.com/aws/amazon-vpc-cni-k8s/pull/1352 is merged

aws/fedora-coreos/kubernetes/workers/fcc/worker.yaml
dghubble: CNI plugins are typically added by daemonset sidecars

dghubble:
This branch (from Dec 2020) is only really meant to show containerd is an option in https://github.com/coreos/fedora-coreos-tracker/issues/767. I'm going to close this until more decisions have been made. So implementation details don't get involved in the core issue of https://github.com/coreos/fedora-coreos-tracker/issues/767, improving container runtime setup on FCOS.

Also, this project doesn't use amazon-vpc-cni

PR #1007, opened 2021-06-25
Prevent NetworkManager from interfering with Calico interfaces

wkruse:
Containers are losing network connectivity, when NetworkManager tries to manage the Calico `cali*` network interfaces. 

```
nmcli
...
cali1a81d481534: connecting (getting IP configuration) to Wired connection 3
        "cali1a81d481534"
        ethernet (veth), EE:EE:EE:EE:EE:EE, sw, mtu 1480
...
```

The fix is to add `/etc/NetworkManager/conf.d/calico.conf` as described in:

https://docs.projectcalico.org/maintenance/troubleshoot/troubleshooting#configure-networkmanager

After the fix and `systemctl restart NetworkManager`:

```
cali1a81d481534: unmanaged
        "cali1a81d481534"
        ethernet (veth), EE:EE:EE:EE:EE:EE, sw, mtu 1480
```

### DIFF
--- a/bare-metal/fedora-coreos/kubernetes/fcc/controller.yaml
+++ b/bare-metal/fedora-coreos/kubernetes/fcc/controller.yaml
@@ -178,6 +178,11 @@ storage:
              echo "Retry applying manifests"
              sleep 5
           done
+    - path: /etc/NetworkManager/conf.d/calico.conf
+      contents:
+        inline: |
+          [keyfile]
+          unmanaged-devices=interface-name:cali*;interface-name:tunl*;interface-name:vxlan.calico
     - path: /etc/sysctl.d/max-user-watches.conf
       contents:
         inline: |
--- a/bare-metal/fedora-coreos/kubernetes/fcc/worker.yaml
+++ b/bare-metal/fedora-coreos/kubernetes/fcc/worker.yaml
@@ -97,6 +97,11 @@ storage:
       contents:
         inline:
           ${domain_name}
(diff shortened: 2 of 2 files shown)


dghubble:
Long ago this used to be needed on Fedora Atomic (see https://github.com/poseidon/typhoon/commit/567e18f015bf7d8e1f111084a04e2d531a2d4a7d), but it no longer should be. On current clusters, these interfaces are ignored already without this change. Is suspect something else is at play in your setup.

```
tunl0: disconnected
        "tunl0"
        iptunnel (ipip), 00:00:00:00, sw, mtu 1440

cali1f95c111389: unmanaged
        "cali1f95c111389"
        ethernet (veth), EE:EE:EE:EE:EE:EE, sw, mtu 1440
```

wkruse:
We didn't see this issue before upgrading to **F34**. Also, it is not always the case, but appears sporadically when some action happens in the cluster (node restarts, drains, etc.). When this happens, NetworkManager grabs **one** of many interfaces and tries to manage it. We didn't change anything related to networking from **F33** to **F34**. The workaround with `calico.conf` fixes it for us. When we remove the fix, NetworkManager tries to manage **all** of the `cali*` interfaces.

We will keep digging.

I'll close the PR.

Do you know, why the fix is no longer needed in FCOS? (Maybe something changed there from **F33** to **F34**.)

wkruse:
Maybe related to https://github.com/coreos/fedora-coreos-tracker/issues/881.

PR #1051, opened 2021-10-06
Add file to hold nodes on iptables-legacy

dghubble:
* Add `/etc/fedora-coreos/iptables-legacy.stamp` to declare that `iptables-legacy` should be used instead of `iptables-nft` (until support is added in future releases)
* https://github.com/coreos/fedora-coreos-tracker/issues/676


dghubble:
Merged as 443bd5a26b56e4a92ed25c798ca2b854752a2844

PR #1072, opened 2021-12-07
Switch FCOS workers to official Fedora CoreOS AMIs

dghubble:
* Fix worker nodes to use official Fedora CoreOS AMIs, instead of the older Poseidon built AMIs (now removed).
This should have been part of #1038, but was missed in code review
* Poseidon build AMIs have been deleted (so I don't have to keep paying to host them for people)


dghubble:
This only related to ARM64. AMD64 was never affected

PR #1102, opened 2022-01-14
Add experimental Flatcar Linux arm64 support on AWS

dghubble:
* Add `arch` variable to Flatcar Linux AWS `kubernetes` and `workers` modules. Accept `amd64` (default) or `arm64` to support
native arm64/aarch64 clusters or mixed/hybrid clusters with arm64 workers
* Requires `flannel` or `cilium` CNI

Similar to https://github.com/poseidon/typhoon/pull/875


dghubble:
Merged as beb9f1477a1e34dd207aa0867f4d6ba47cb9bff2

PR #1114, opened 2022-02-07
Change default CNI provider from Calico to Cilium

dghubble:
* Cilium (v1.8) was added to Typhoon in v1.18.5 in June 2020 and its become more impressive since then. Its currently the leading CNI provider choice.
* Calico has grown complex, has lots of CRDs, masks management complexity with an operator (which we won't use), doesn't provide multi-arch images, and hasn't been compatible with Kubernetes v1.23 (with ipvs) for several releases.
* Both have CNCF conformance quirks (flannel used for conformance), but that's not the main factor in choosing the default

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@ Notable changes between versions.
 
 ## Latest
 
+* Change default CNI `networking` provider from `calico` to `cilium` ([#1114](https://github.com/poseidon/typhoon/pull/1114))
+
 ## V1.23.3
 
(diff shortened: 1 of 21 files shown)


xmulligan:
@dghubble glad to hear that you love Cilium and see it as the leading CNI 🚀 

would you mind adding Typhoon to the Cilium adopters list? https://github.com/cilium/cilium/blob/master/USERS.md

We need this for when we apply to graduate at the CNCF

PR #1158, opened 2022-05-01
Add Better AWS Worker Tagging

fifofonix:
Propagate any AWS Terraform Provider Default Tags To Worker Instances & Volumes for better AWS account management.

The only tag attached to worker nodes presently when they are spun up by the ASG is their `Name`.  In many scenarios however it is desirable to classify infrastructure with additional tags whether it be for billing / cost analysis / environment segregation etc.  Indeed the AWS Terraform Provider provides for this in its declaration.  This works out-of-the-box for many assets, e.g. in Typhoon's case for EC2 controller nodes, NLBs etc.  However, Application Scaling Groups need special attention if the intent is to have the dynamic nodes they spin up tagged correctly.  This enhancement addresses this point. 

Doing so has necessitated switching from AWS's legacy Launch Config methodology to the newer recommended Launch Template methodology which has more flexibility in amongst other things specifying tags.

* Change is confined solely to fedora-coreos `workers.tf` presently pending maintainer feedback.
* Change will be extended to cover flatcar too if there is buy-in for this change.

## Testing

* Terraform used to update existing spot worker node cluster in-place successfully and new worker node EC2/EBS tags validated via console.
* New terraform cluster deployed with on-demand worker nodes and tags validated on worker nodes.

## Launch Template Items For Discussion

* Launch Template Root block device declarations require a device name which has been hard coded to `/dev/xvda` which should work fine for Fedora CoreOS x86 AMIs ... for the time being ... but is perhaps potentially brittle.  Also, aarch64?
* Launch Templates are versioned objects in AWS.  We define the ASG to use the `$Latest` which seems reasonable given the Launch Template is not shared.
* Launch Templates with spot instances require an `instance_market_options` block which seemingly does not support on-demand instances.  Instead for on-demand the block is absent.  A dynamic terraform block implements this.

## Other Topics For Discussion

* Right now I've arbitrarily chosen to apply/cascade tagging to `instance` and `volume` assets.  It is equally possible to also cascade to other assets, e.g. `elastic-gpu, network-interface, spot-instances-request`.  See [here](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateTagSpecificationRequest.html).

### DIFF
--- a/aws/fedora-coreos/kubernetes/workers/workers.tf
+++ b/aws/fedora-coreos/kubernetes/workers/workers.tf
@@ -1,6 +1,6 @@
 # Workers AutoScaling Group
 resource "aws_autoscaling_group" "workers" {
-  name = "${var.name}-worker ${aws_launch_configuration.worker.name}"
+  name = "${var.name}-worker ${aws_launch_template.worker.name}"
 
   # count
   desired_capacity          = var.worker_count
@@ -13,7 +13,10 @@ resource "aws_autoscaling_group" "workers" {
   vpc_zone_identifier = var.subnet_ids
 
   # template
-  launch_configuration = aws_launch_configuration.worker.name
+  launch_template {
+      id      = aws_launch_template.worker.id
+      version = "$Latest"
+    }
 
   # target groups to which instances should be added
   target_group_arns = flatten([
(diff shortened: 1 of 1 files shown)


fifofonix:
@dhubble this is my first pull request so apologies for any unintentional faux-pas, e.g. if it is bad form to submit one without an accompanying issue.  Do you have any thoughts on content?  (Awesome project by the way - I'm inching towards using it in anger but currently just experimenting with two small test clusters).

fifofonix:
No problem.  Closing.

PR #1174, opened 2022-06-08
Adding AAAA alias records for NLB on AWS

### LINKED ISSUE #1173: AWS worker nodes not being auto-assigned an IPv4 address
Typhoon deployed with the following configuration (barring the `xxxx` placeholders) while reading the documentation [here](https://typhoon.psdn.io/fedora-coreos/aws/):

```
module "kubernetes" {
  source             = "git::https://github.com/poseidon/typhoon//aws/fedora-coreos/kubernetes?ref=v1.24.1"
  cluster_name       = "kubernetes"
  dns_zone           = aws_route53_zone.xxxx.name
  dns_zone_id        = aws_route53_zone.xxxx.id
  ssh_authorized_key = "ssh-rsa xxxx"
  worker_count       = 2
  network_mtu        = 8981
}
```

The output of `kubectl get nodes` lists only the control node. On investigating, the worker nodes are failing to reach the url `https://kubernetes.xxxx.com:6443/` and are experiencing timeouts. This is because the worker nodes have no public IPv4 addresses and there is no NAT Gateway to proxy requests.

Adding a AAAA record that otherwise matches the root A record (alias to the NLB) fixes the issue.

JamesBelchamber:
Adds an AAAA alias record pointing to the NLB for AWS deployments with CoreOS and Flatcar.

Fixes #1173

### DIFF
--- a/aws/fedora-coreos/kubernetes/nlb.tf
+++ b/aws/fedora-coreos/kubernetes/nlb.tf
@@ -1,9 +1,24 @@
-# Network Load Balancer DNS Record
+# Network Load Balancer DNS A Record
 resource "aws_route53_record" "apiserver" {
   zone_id = var.dns_zone_id
 
   name = format("%s.%s.", var.cluster_name, var.dns_zone)
   type = "A"
+
+  # AWS recommends their special "alias" records for NLBs
+  alias {
+    name                   = aws_lb.nlb.dns_name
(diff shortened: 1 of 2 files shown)


dghubble:
I'll followup depending on how Terraform AWS addresses the root issue.

PR #1207, opened 2022-08-14
Roll instance template changes to worker managed instance groups

dghubble:
* When a worker managed instance group's (MIG) instance template changes (including `worker_type`, `disk_size`, `preemptible`, or Butane `worker_snippets` but excluding new AMIs), use Google Cloud's rolling update features to ensure instances match declared state
* Ignore new AMIs since Fedora CoreOS and Flatcar Linux nodes already auto-update and reboot themselves
* Rolling updates will create surge instances, wait for health checks, then delete old instances (0 unavilable instances)
* Instances are replaced to ensure new Ignition/Butane snippets are respected
* Add managed instance group autohealing (i.e. health checks) to ensure new instances' Kubelet is running

Renames

* Name apiserver and kubelet health checks consistently
* Rename MIG from `${var.name}-worker-group` to `${var.name}-worker`

Rel: https://cloud.google.com/compute/docs/instance-groups/rolling-out-updates-to-managed-instance-groups

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,11 +32,23 @@ version: 1.0.0
 
 ### AWS
 
-* Rename worker autoscaling group `${cluster_name}-worker`
-* Rename launch configuration `${cluster_name}-worker`
+* Rename worker autoscaling group `${cluster_name}-worker` ([#1202](https://github.com/poseidon/typhoon/pull/1202))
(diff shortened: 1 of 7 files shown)


dghubble:
May change this to health check kube-proxy instead of Kubelet before release

PR #1286, opened 2023-01-04
Support using a different install disk target per node on baremetal

woneill:
Support using a different install disk target per node during baremetal installations.
In my use case, my controller node uses /dev/sda but the worker nodes use /dev/nvme0n1

* Breaking changes: deprecate `var.install_disk` and require Terraform >= 1.3.0.
  * The Terraform version is to support the [optional object type attribute](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#optional-object-type-attributes) that now holds the install disk target value.
* Refactoring to resolve `terraform_deprecated_index` warnings from `tflint`
  * Legacy "attribute-only" splat expressions have been deprecated for a while: https://developer.hashicorp.com/terraform/language/expressions/splat#legacy-attribute-only-splat-expressions

## Testing

I've installed my own baremetal cluster (flatcar linux) using this branch

### DIFF
--- a/bare-metal/fedora-coreos/kubernetes/bootstrap.tf
+++ b/bare-metal/fedora-coreos/kubernetes/bootstrap.tf
@@ -4,7 +4,7 @@ module "bootstrap" {
 
   cluster_name                    = var.cluster_name
   api_servers                     = [var.k8s_domain_name]
-  etcd_servers                    = var.controllers.*.domain
+  etcd_servers                    = var.controllers[*].domain
   networking                      = var.networking
   network_mtu                     = var.network_mtu
   network_ip_autodetection_method = var.network_ip_autodetection_method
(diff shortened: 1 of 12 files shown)


dghubble:
This'll have to wait a while. Its too soon to be dropping support for Terraform v1.2 and when we do there will be a broader set of refactors.

dghubble:
I'm planning to tackle this in a different way. There are a number of variables that can't currently be defined on a per-machine basis. For some time now, I've considered a restructuring that would address the general issue and any future per-variable, modeled on what we do for cloud worker pools. As a plus, it doesn't require any new Terraform features and supports both the existing style (no breaking changes) and a new per-machine style.

https://github.com/poseidon/typhoon/pull/1295 is a draft of what this would look like. I believe it addresses your case.

woneill:
Nice! I like that worker approach much better.

PR #1302, opened 2023-02-27
Expose flatcar-install OEM parameter

MLNW:
By exposing this parameter it is possible to install OEM specific software during the `flatcar-install` invocation.

## Testing

I installed a Kubernetes cluster on vSphere VMs and used the newly exposed flag to install the vmware OEM files. Now vSphere no longer complains that the vSphere tools are not installed and it is possible to see the assigned hostname and IPs in the vSphere UI. This also enables the vSphere Terraform provider to wait for IP assignment.

When using `cached_install = true` Matchbox also needs to serve these files otherwise the `installer.service` fails:
```
https://${channel}.release.flatcar-linux.net/amd64-usr/${version}/flatcar_production_${oem_type}_image.bin.bz2
https://${channel}.release.flatcar-linux.net/amd64-usr/${version}/flatcar_production_${oem_type}_image.bin.bz2.sig
```

### DIFF
--- a/bare-metal/flatcar-linux/kubernetes/profiles.tf
+++ b/bare-metal/flatcar-linux/kubernetes/profiles.tf
@@ -55,6 +55,7 @@ data "ct_config" "install" {
     mac                = concat(var.controllers.*.mac, var.workers.*.mac)[count.index]
     install_disk       = var.install_disk
     ssh_authorized_key = var.ssh_authorized_key
+    oem_type           = var.oem_type
     # only cached profile adds -b baseurl
     baseurl_flag = var.cached_install ? "-b ${var.matchbox_http_endpoint}/assets/flatcar" : ""
   })
--- a/bare-metal/flatcar-linux/kubernetes/variables.tf
+++ b/bare-metal/flatcar-linux/kubernetes/variables.tf
@@ -156,6 +156,17 @@ variable "enable_aggregation" {
   default     = true
 }
 
+variable "oem_type" {
+  type        = string
+  description = <<EOD
+An OEM type to install with flatcar-install. Find available types by looking for Flatcar image files
+ending in `image.bin.bz2`. The OEM identifier is contained in the filename.
+E.g., `flatcar_production_vmware_raw_image.bin.bz2` leads to `vmware_raw`.
+See: https://www.flatcar.org/docs/latest/installing/bare-metal/installing-to-disk/#choose-a-channel
+EOD
+  default     = ""
+}
+
 # unofficial, undocumented, unsupported
 
 variable "cluster_domain_suffix" {
--- a/bare-metal/flatcar-linux/kubernetes/worker/matchbox.tf
+++ b/bare-metal/flatcar-linux/kubernetes/worker/matchbox.tf
@@ -50,6 +50,7 @@ data "ct_config" "install" {
     mac                = var.mac
     install_disk       = var.install_disk
     ssh_authorized_key = var.ssh_authorized_key
+    oem_type           = var.oem_type
     # only cached profile adds -b baseurl
     baseurl_flag = var.cached_install ? "-b ${var.matchbox_http_endpoint}/assets/flatcar" : ""
   })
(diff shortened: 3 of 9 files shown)


MLNW:
@dghubble the last part the description above complicates the merge of this PR. I would propose adding an option to the [matchbox/get-flatcar](https://github.com/poseidon/matchbox/blob/main/scripts/get-flatcar) script that downloads all provided `oem_type` files. Then users may deploy Matchbox with a supported set of `oem_types` which may be used in conjunction with the newly exposed flag when deploying a baremetal cluster with Typhoon. If you agree I will create a PR for that as well.

bare-metal/flatcar-linux/kubernetes/worker/matchbox.tf
```diff
@@ -50,6 +50,7 @@ data "ct_config" "install" {
     mac                = var.mac
     install_disk       = var.install_disk
     ssh_authorized_key = var.ssh_authorized_key
+    oem_type = var.oem_type
```
dghubble, line 53: terraform fmt

bare-metal/flatcar-linux/kubernetes/variables.tf
```diff
@@ -155,6 +155,12 @@ variable "enable_aggregation" {
   default     = true
 }
 
+variable "oem_type" {
+  type        = string
+  default     = ""
```
dghubble, line 160: not: default as the last assignment to match the other variable blocks please

bare-metal/flatcar-linux/kubernetes/variables.tf
```diff
@@ -155,6 +155,12 @@ variable "enable_aggregation" {
...
+  description = "An OEM type to install with flatcar-install."
```
dghubble, line 161: Can you include the list of valid Flatcar OEMs in the description, like `(e.g. a, b, c)` and link to Flatcar Linux docs

bare-metal/flatcar-linux/kubernetes/profiles.tf
```diff
@@ -55,6 +55,7 @@ data "ct_config" "install" {
     mac                = concat(var.controllers.*.mac, var.workers.*.mac)[count.index]
     install_disk       = var.install_disk
     ssh_authorized_key = var.ssh_authorized_key
+    oem_type = var.oem_type
```
dghubble, line 58: fmt

dghubble (commented):
Setting the flatcar-linux OEM flag is ok. I don't use VMWare or the others so I won't validate them, but this change can just be oriented around allowing the parameter to be set. Its up to Flatcar Linux what that does, since I don't want OEM specific features creeping into the distro itself.

Can you also dig up more info from the Flatcar Linux folks about the OEMs they support / will support?

dghubble:
I don't think the Matchbox script needs an update. Folks wanting OEM specific stuff can download it.

MLNW:
> I don't think the Matchbox script needs an update. Folks wanting OEM specific stuff can download it.

We use a systemd unit like this:
```yaml
    - name: "download-flatcar.service"
      enabled: true
      contents: |
        [Unit]
        Description=Download Flatcar image
        Wants=systemd-resolved.service
        After=clone-matchbox.service
        Before=matchbox.service

        [Service]
        Type=oneshot
        RemainAfterExit=true

        ExecStart=/usr/bin/bash -c '/usr/bin/test -d /opt/matchbox.opt/matchbox/assets || /opt/matchbox/scripts/get-flatcar ${flatcar_channel} ${flatcar_version} /opt/matchbox.opt/matchbox/assets'

        # WORKAROUND: Needed to support installing VMWare tools
        ExecStartPost=/usr/bin/curl \
          --output-dir /opt/matchbox.opt/matchbox/assets/flatcar/${flatcar_version} \
          -O https://${flatcar_channel}.release.flatcar-linux.net/amd64-usr/${flatcar_version}/flatcar_production_vmware_raw_image.bin.bz2
        ExecStartPost=/usr/bin/curl \
          --output-dir /opt/matchbox.opt/matchbox/assets/flatcar/${flatcar_version} \
          -O https://${flatcar_channel}.release.flatcar-linux.net/amd64-usr/${flatcar_version}/flatcar_production_vmware_raw_image.bin.bz2.sig

        [Install]
        WantedBy=matchbox.service

```

bare-metal/flatcar-linux/kubernetes/variables.tf
MLNW: I wrote some more documentation here. PTAL and let me know if its okay.

PR #1306, opened 2023-03-03
Add support for devince binding

puppetmaster:
In some cases we need to use block devices directly (Physical volume storage).

This new variable allows the user to mount devices with the --device option for flatcar baremetal.

How to use :

add a list of objects like this to the cluster definition :
```
worker_bind_devices = [
  { source = "/dev/sdb" target = "/dev/sdb" },
  { source = "/dev/sdc" target = "/dev/sdd" },
  { source = "/dev/sdd" target = "/dev/sdd" }
]
```
This will bind the source device on the target directory in evry node.

* Specific change
* Specific change

## Testing

### DIFF
--- a/bare-metal/flatcar-linux/kubernetes/variables.tf
+++ b/bare-metal/flatcar-linux/kubernetes/variables.tf
@@ -73,6 +73,20 @@ variable "worker_node_taints" {
   default     = {}
 }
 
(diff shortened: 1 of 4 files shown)


puppetmaster:
This is not ready, sorry for the noise

Issue #1026, opened 2021-08-05
[Proposal] Sensitive info on matchbox only for designated target


torrescd:
I don't know if this is the correct place to share this thought.
--------------------

If matchbox is able to share in a safe way information only with the target node, we could avoid a lot of complexity in typhoon post-constructs via remote-exec (and ssh problems as local/trigger machine won't need to connect with node)

Regards!

dghubble:
PXE boot enables unprovisioned machines to provision themselves from the network very early in their lifecycle. There is no root identity or basis for authenticating that a client node is who is claims. Matchbox should not be used as a store for sensitive information. Machine attestation is a complex topic, and there is a reason we use SSH to distribute materials securely.

https://matchbox.psdn.io/deployment/#operational-notes

torrescd:
I see,

And how about serving that information only once and then erase it?

dghubble:
That's not a security mechanism either. I'll leave you with the answer I gave someone in 2018 regarding early boot trust. 

> You should not store sensitive information in Matchbox. By definition, its one of the earliest configs a bare-metal machine loads. That's why you'll see Typhoon and Tectonic copy sensitive information onto the node after bring-up. That's why you see the examples do it this way and the reason for the recommendation. Place your SSH public key on the box and copy secrets. Or fetch them with units during startup, from some other attesting store (just shifts the problem).

> Many wonder "why not just add authentication"? But the client machine must authenticate with something, some information. How did that something get there? Matchbox is the earliest service. You can pick an even earlier service to add an auth token/cert/key into, but then you have the same question - how did the "fresh from the manufacturer" server get a key? Eventually, you'll end up using the TPM and collecting keys upfront or other forms of strong attestation for physical machine identity.

> On cloud providers, folks can dodge the problem, since you'd inherently trust the cloud provider and whatever secret distribution mechanism they offer.

> Perhaps more than you wanted to know. But bootstrapping from zero state to trust is not an easy problem. As you've guessed.

torrescd:
Thank you for your answer. It's an interesting topic.

One of my concerns over the first approach I said is:

 Why when you ssh to your node you can trust* that it is him (DNS/IP) but matchbox cannot trust that origin IP as safe.
* (Public cert can be been using it by anyone else as matchbox expose it)

If the network is compromised, won't be those two operations equally unsafe?

Regards!

dghubble:
[...]

PR #1320, opened 2023-03-28
os_architecture option

### LINKED ISSUE #1319: Adding support for ARM64 architecture on bare-metal with flatcar-linux
### Overview

As of now, flatcar-linux on bare-metal only supports the amd64 architecture since the "amd64-usr" string is hardcoded everywhere in the code. I would like to change this and create an optional variable, which would default to amd64, to let an user choose his architecture and make typhoon with flatcar-linux on bare metal compatible with the aarch64 architecture. 

### Motivation

I am currently working on an aarch64 cluster and would like to make some changes which would make typhoon with flatcar-linux on bare-metal compatible with the aarch64 architecture.

Sadi-a:
Until now, typhoon with flatcar-linux on bare metal was only available for machines with an AMD64 (x86_64) architecture. This means there is no currently no support of ARM64 (aarch64) machines. In this commit, the support for different architectures on flatcar-linux with bare-metal was added. This was done through the addition of an os_architecture variable which replaces the hardcoding of the "amd64" string in some places, such as in the links to the flatcar-linux releases

Fixes #1319

### DIFF
--- a/bare-metal/flatcar-linux/kubernetes/butane/install.yaml
+++ b/bare-metal/flatcar-linux/kubernetes/butane/install.yaml
@@ -35,6 +35,7 @@ storage:
             -d ${install_disk} \
             -C ${os_channel} \
             -V ${os_version} \
+            -B ${os_architecture}-usr \
             ${baseurl_flag} \
             -i ignition.json
           udevadm settle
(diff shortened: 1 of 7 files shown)


dghubble:
See https://github.com/poseidon/typhoon/issues/1026#issuecomment-893663674

PR #1321, opened 2023-03-28
Generic selector baremetal flatcar

### LINKED ISSUE #1317: Adding an "extra" variable for machine identification
### Overview

This issue concerns flatcar-linux on bare-metal. 
As of now, the selector only uses the MAC address. It might be useful to be able to use more fields in it, for example a machine's UUID or its CPU architecture. I could implement this change through a data structure such as a map. 

### Motivation

This change would give more freedom of movement when it comes to retrieving ignition.json files from Matchbox.

Sadi-a:
This Pull request concerns flatcar-linux on bare metal. It aims to add a generic selector which could be used with matchbox to use some other selectors than just the MAC address, such as a machine's architecture or UUID for example. 
This is done here through the addition of an 'extra_selectors' variable which uses the 'map' data structure and defaults on {}.

Fixes #1317

### DIFF
--- a/bare-metal/flatcar-linux/kubernetes/butane/install.yaml
+++ b/bare-metal/flatcar-linux/kubernetes/butane/install.yaml
@@ -30,7 +30,7 @@ storage:
       contents:
         inline: |
           #!/bin/bash -ex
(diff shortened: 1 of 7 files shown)


dghubble:
Thanks, but no thanks.

PR #1322, opened 2023-03-28
Avoiding having to launch SSH after a node's reboot

### LINKED ISSUE #1316: Avoiding having to run terraform post boot.
### Description 

Currently, Typhoon uses SSH to give the workers and controllers some secrets and other files like certificates for example. 
Thus, I would like to make it so SSH is not needed to give secrets and other required files with flatcar-linux on bare metal. One way to do this would be to add an option which would allow one to include these files in the ignition.json file's definition. That way, SSH wouldn't be needed anymore and only the matchbox server would work on it.

### Motivation

When a node reboots, a human intervention is required to run `terraform apply`. With the addition of this feature, it wouldn't be necessary to run this command and the node would be ready immediately without waiting for a `terraform apply`.

Sadi-a:
Currently, `terraform apply` has to be run after the nodes have booted, which means that if one node reboots, human intervention will be required to run `terraform apply` again, or else the node won't be instanced and will stay empty, without the information necessary for it to join the cluster again.
This commit aims to change that through the implementation of an "enable_SSH" option, which defaults to true. Disabling this option will put the files originally distributed through SSH in the butane file, which means they will be included in the ignition.json file and so the machine will go get them just like the other instructions in the ignition file.

Fixes #1316

### DIFF
--- a/bare-metal/flatcar-linux/kubernetes/butane/controller.yaml
+++ b/bare-metal/flatcar-linux/kubernetes/butane/controller.yaml
@@ -124,6 +124,18 @@ systemd:
         ExecStartPost=/bin/touch /opt/bootstrap/bootstrap.done
         [Install]
         WantedBy=multi-user.target
+    - name: bootstrap-layout.service
+      contents: |
+        [Unit]
+        Description=Kubernetes layout
+        Wants=bootstrap.service
+        After=bootstrap.service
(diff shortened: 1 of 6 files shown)


dghubble:
See https://github.com/poseidon/typhoon/issues/1026#issuecomment-893663674

PR #1335, opened 2023-04-24
Fix: duplicate kernel_args added to bare-metal workers

woneill:

* `var.kernel_args` is already included in `local.args` so concat-ing it again results in duplicate arguments on bare-metal workers

## Testing

Noted that passed in `kernel_args` were showing up twice in terraform plan output


dghubble:
Thanks!

PR #1366, opened 2023-08-03
Update docs on flatcar-linux bare-metal kubernetes worker module usage.

puppetmaster:
Correctif typos in flatcar-linux bare-metal kubernetes worker module doc.

* code example corrections (typo)

## Testing

Run the doc, and you will see, that now it quite works.


dghubble (changes requested):
These are not docs fixes, you're including personal features that are not part of any other mainline modules

Issue #1386, opened 2023-10-13
Bare-metal flatcar is missing "kubeconfig" output required for separate worker module


acmcarther:
**Description**

The [flatcar linux baremetal workflow](https://github.com/poseidon/typhoon/blob/main/docs/flatcar-linux/bare-metal.md) describes the option of  "discrete workers that attach to the cluster". This requires creating a separate module for the worker and setting the worker's kubeconfig to `module.cluster-name.kubeconfig`. When trying to do this though, during `terraform plan`, I get this:

```
$ terraform plan
╷
│ Error: Unsupported attribute
│
│   on cluster.tf line 58, in module "dominion-big-worker-1":
│   58:   kubeconfig = module.dominion.kubeconfig
│     ├────────────────
│     │ module.dominion is a object
│
│ This object does not have an attribute named "kubeconfig".
╵
```

That output is clearly missing from [flatcar's bare-metal outputs](https://github.com/poseidon/typhoon/blob/6bd6d46fb28849821456c0c69c978cfa69d8bbbc/bare-metal/flatcar-linux/kubernetes/outputs.tf). You can see it in the [corresponding fedora-coreos outputs though](https://github.com/poseidon/typhoon/blob/6bd6d46fb28849821456c0c69c978cfa69d8bbbc/bare-metal/fedora-coreos/kubernetes/outputs.tf#L6-L11)

My [cluster.tf is here](https://gist.github.com/acmcarther/2a1a98cf8766d6d595bf8415addd1a7e)

**Steps to Reproduce**

1. Create a cluster.tf with standard bare-metal options for flatcar linux
2. Add another module for worker, attempting to reference the cluster's kubeconfig.

**Expected behavior**

No error, worker in separate module uses kubeconfig from cluster

**Environment**

* Platform: bare-metal
* OS: flatcar-linux
* Release: v1.28.2
* Terraform and plugins:
```
terraform version
Terraform v1.6.1
on darwin_arm64
+ provider registry.terraform.io/hashicorp/local v2.4.0
+ provider registry.terraform.io/hashicorp/null v3.2.1
+ provider registry.terraform.io/hashicorp/random v3.5.1
+ provider registry.terraform.io/hashicorp/tls v3.4.0
+ provider registry.terraform.io/poseidon/ct v0.13.0
+ provider registry.terraform.io/poseidon/matchbox v0.5.2
```

**Possible Solution**

Add the kubeconfig output to the bare-metal flatcar outputs.tf

[...]

PR #1387, opened 2023-10-13
Update outputs.tf for bare-metal/flatcar-linux to include kubeconfig output

acmcarther:
This PR adds the `kubeconfig` output to the bare-metal/flatcar-linux outputs.tf in order to fix support for separately defined worker modules.

Before this change, that functionality was broken as described in https://github.com/poseidon/typhoon/issues/1386

## Testing

`terraform plan` using my patched outputs.tf completes.

rel: 1386


dghubble:
Thanks

Issue #1350, opened 2023-06-06
Cannot run AWS nginx ingress controler addon


cosad3s:
**Description**

The nginx ingress controller addon does not launch.

**Steps to Reproduce**

- Deploy fresh Typhoon cluster in **AWS** with **Calico CNI** (`networking   = "calico"`)
- Add nginx ingress controller addon : `kubectl apply -R -f addons/nginx-ingress/aws` (Followed: [https://typhoon.psdn.io/addons/ingress/](https://typhoon.psdn.io/addons/ingress/))
- We got crashes on the pods:

```bash
❯ kubectl get pods --all-namespaces
NAMESPACE     NAME                                       READY   STATUS             RESTARTS      AGE
ingress       nginx-ingress-controller-c966f5589-h7mn2   0/1     CrashLoopBackOff   6 (17s ago)   4m58s
ingress       nginx-ingress-controller-c966f5589-t8r6t   0/1     CrashLoopBackOff   6 (7s ago)    4m58s
```

**Logs**

```bash
❯ kubectl logs nginx-ingress-controller-c966f5589-h7mn2 -n ingress
-------------------------------------------------------------------------------
NGINX Ingress controller
  Release:       v1.5.1
  Build:         d003aae913cc25f375deb74f898c7f3c65c06f05
  Repository:    https://github.com/kubernetes/ingress-nginx
  nginx version: nginx/1.21.6

-------------------------------------------------------------------------------

W0606 14:27:31.641402       7 client_config.go:617] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0606 14:27:31.641700       7 main.go:209] "Creating API client" host="https://10.3.0.1:443"
```

**Expected behavior**

nginx ingress controller AWS addon does not crash

**Environment**

* Platform: AWS
* OS: fedora-coreos
* Release: Typhoon 1.27.2 
* Terraform: Terraform v1.4.6
* Plugins: poseidon/ct 0.13.0, hashicorp/aws 4.61.0

dghubble:
I've been using Terraform to manage these Kubernetes resources across my clusters, and no longer update these raw Kubernetes manifests examples in this repo. The addons should probably be removed.  I think early on they were important to show, but these days people can probably run their own Nginx, Prometheus, etc. without me maintaining examples.

dghubble:
[...]

PR #1409, opened 2024-01-12
Fix/Ingress on gcp doesnt work by default.

8ball030:

When deployed to GCP I hit this issue;

 k8s.io/client-go@v0.25.3/tools/cache/reflector.go:169: Failed to watch *v1.EndpointSlice: failed to list *v1.EndpointSlice: endpointslices.discovery.k8s.io is forbidden: User "system:serviceaccount:ingress:default" cannot list resource "endpointslices" in API group "discovery.k8s.io" at the cluster scope 

I added the extra cluster rbacs to the ingress class and all seems to now work.

## Testing

I deployed on GCP and was able to route ingress traffic to a contianer in the cluster.

### DIFF
--- a/addons/nginx-ingress/google-cloud/rbac/cluster-role.yaml
+++ b/addons/nginx-ingress/google-cloud/rbac/cluster-role.yaml
@@ -59,4 +59,11 @@ rules:
       - get
       - list
       - watch
-
+  - apiGroups: 
+      - discovery.k8s.io
+    resources:
+      - "endpointslices"
+    verbs: 
+      - get
+      - list
+      - watch


dghubble:
Thanks

Closes #1350

I'd be curious to hear about your usage because I've been weighing removing the addon examples Kubernetes manifests, or maybe replacing them with Terraform modules.

8ball030:
Oh that was convenient. 
Wasn't aware I'd fixed an issue. 
I am a pretty heavy user of this to be candid. 
It is kind of my go to. 

In terms of improvements, I always end up having to basically do a couple of things;

A) setup the ingress controller. 
(An option in the initial tf module I.e. ingress_enabled: bool) would save me time. 

B) normally I end up having to setup a wildcard dns entry into the cluster 

C) I normally have to deploy certmanager 

D) I almost always have to setup persistent storage. For example an efs mount to the workers such that I can use a nfs controller to manage volumes. 

Thanks again for merging my pr, it's quite cool to be a contributor to this repo, I've been using it for years.

PR #1419, opened 2024-02-24
Allow CNI networking to be set to none

dghubble:
* Set CNI networking to "none" to skip installing any CNI provider (i.e. no flannel, Calico, or Cilium). In this mode, cluster nodes will be NotReady until you add your own CNI stack
* Motivation: I now tend to manage CNI components as addon modules just like other applications overlaid onto a cluster. It allows for faster iteration and may eventually become the recommendation

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ Notable changes between versions.
 
 * Kubernetes [v1.29.2](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.29.md#v1292)
 * Update flannel from v0.22.2 to [v0.24.2](https://github.com/flannel-io/flannel/releases/tag/v0.24.2)
+* Allow CNI `networking` to be set to "none" to skip bootstrapping flannel, Calico, or Cilium ([#1419](https://github.com/poseidon/typhoon/pull/1419))
 
 ## v1.29.1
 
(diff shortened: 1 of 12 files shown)


kalmufti:
Thanks! Been looking forward to this feature. 

Would this be the best way to use it:
```terraform
networking = "none"
```

Or through the new variable from #1421? I see you included the change from #1421 in the release notes but not this #1419. 

Personally I like setting up this way as it looks cleaner, and since you already checking the state  of both ```var.networking : "none"``` and the new ```install_container_networking``` variable in ```kubernetes/bootstrap.tf```

dghubble:
You should set `install_container_networking` to false if you want to manage the flannel, Calico, or Cilium Kubernetes resources yourself. You will still need to have `networking` set to the corresponding CNI provider to configure cloud firewall / security group rules or other conditionals. The release notes guide you this direction.

kalmufti:
Are you saying the ```networking``` variable is still important to be set for cloud providers only? I'm deploying on bare-metal clusters. I've tested 2 clusters by just setting ```networking = "none"``` and it worked as expected, meaning it deployed without CNI and was able to deploy my Cilium setup as I needed.

Looking at the ```<platform>/fedora-coreos/kubernetes/bootstrap.tf``` files I see this condition.
```terraform
networking = var.install_container_networking ? var.networking : "none"
```
i.e what I did by setting ```networking = "none"``` it eventually achieved what I'm after.
```terraform
if var.install_container_networking == true:
  networking = "CNI" # in my case I set it to "none"
else:
  networking = "none"
```
So wouldn't ```var.install_container_networking``` be redundant in this case, unless ```var.networking``` is still needed elsewhere?

dghubble:
The two options serve different purposes. It doesn't currently make much difference on bare-metal on these platforms that don't use host-level firewall rules but may in future.

Continue to use them as I've described

PR #1424, opened 2024-02-27
Update controllers.tf

8ball030:
Allows the controllers to be updated and scaled.

At present, if you try to scale the controllers, you have a bad time, as the capacity type is immutable.

However with this change, you can change the controller type and it will;

stop the controller;

change the capacity/size

start the controller.

I ran this on a deployment w 2 controllers and all seemed to be ok.

Was hoping you could get this merged :)


dghubble (commented):
Can you update both Fedora CoreOS and Flatcar Linux please?

8ball030:
@dghubble 
This is now actioned, thanks

PR #1511, opened 2024-09-02
feat: add option to enable IMDSv2

### LINKED ISSUE #1510: secure ec2 enabling IMDSv2
**Description**

in some environments is not possible to deploy ec2 instances that run  using IMDSv1

this issue is to request enabling it or at least discuss it.

the change would be something like

  metadata_options {
    http_tokens                 = "required"
    http_put_response_hop_limit = 1
  }

restricting HTTP PUT can have an effect on tje cluster depending how is implemented, if pods are assuming the EC2 instance role.

bruvio:
Adding terraform variables to optionally enable IMDSv2 as in some organizations is not possible to deploy EC2 instances that do not have IMDSv2 enabled.

tested deploying 
```
module "bruvio" {
  source = "git::https://github.com/bruvio/typhoon//aws/flatcar-linux/kubernetes?ref=feature/enable-IMDSv2"



  cluster_name       = var.cluster_name
  dns_zone           = var.dns_zone
  dns_zone_id        = var.dns_zone_id
  ssh_authorized_key = var.ssh_authorized_key
  networking         = "calico"
  network_mtu        = 8981
  components         = local.custom_components


  # optional
  host_cidr          = "10.0.0.0/16"
  controller_count   = 1
  worker_count       = 2
  worker_node_labels = ["worker"]
  http_tokens = "required"
  http_put_response_hop_limit = 1
}
```

resolves #1510

### DIFF
--- a/aws/fedora-coreos/kubernetes/controllers.tf
+++ b/aws/fedora-coreos/kubernetes/controllers.tf
@@ -16,6 +16,10 @@ resource "aws_route53_record" "etcds" {
 # Controller instances
 resource "aws_instance" "controllers" {
   count = var.controller_count
+  metadata_options {
+    http_tokens                 = var.http_tokens
+    http_put_response_hop_limit = var.http_put_response_hop_limit
+  }
 
   tags = {
     Name = "${var.cluster_name}-controller-${count.index}"
--- a/aws/fedora-coreos/kubernetes/variables.tf
+++ b/aws/fedora-coreos/kubernetes/variables.tf
@@ -213,3 +213,13 @@ variable "components" {
   })
   default = null
 }
+
+variable "http_tokens" {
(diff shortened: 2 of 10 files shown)


JordanP:
This looks highly specific to AWS and even for AWS this doesn't look typical. Should typhoon support that ?

dghubble:
Yeah, I don't intent this to be a full variable. If you need it, just set it in your own fork. 

At some point I'll probably switch it to required, but AWS pushing this is somewhat security theater imo and breaks scripts not using AWS's SDK (possibly their point). Changing scripts is ok but annoying. And AWS is the only platform doing this. Here, I'm choosing not to enforce this on people, v1 or v2 should work. https://github.com/poseidon/typhoon/pull/1412

PR #1519, opened 2024-09-24
doc: set file_permission 0600 for kubeconfig file

JordanP:
It's only documentation, but kubeconfig file contains sensitive info so it's better to secure it a little


dghubble (approved):
lgtm

